Issue #31, opened 2025-05-01
Modify `make start` and `make setup` to use celestia-app


Manav-Aggarwal:
Replace the docker container in `make start` and `make setup` from simapp to celestia-app. 

Use this feature branch in celestia-app: https://github.com/celestiaorg/celestia-app/tree/feature/zk-execution-ism

damiannolan:
I have opened #38 for this. Proposing that we use a new docker-compose file and incrementally build it up. We can adjust the make targets later

Issue #1, opened 2025-04-28
Modify `make setup` to setup Hyperlane contracts, ISMs, warp routes


rootulp:
Currently `make setup` creates two IBC light clients and connects them. To adapt this work to Hyperlane, we should instead run `make setup` to deploy Hyperlane contracts and set up two hyperlane warp routes instead from Celestia to the EVM EVM rollup.

The route from Celestia to the EVM rollup should use a Multisig ISM while the route back can either use a Multisig ISM or a NoOp ISM for now whichever is simpler.

Will also need to replace the hardcoded IBC relayer with the current Hyperlane relayer instead.

For this work, the `make start` command shouldn't need to run the following docker images anymore:

celestia-prover
evm-prover
indexer

damiannolan:
```
Now updating chain rethlocal at filesystem registry at /Users/damiannolan/.hyperlane
Done updating chain rethlocal at filesystem registry
✅ Core contract deployments complete:

    staticMerkleRootMultisigIsmFactory: "0x2854CFaC53FCaB6C95E28de8C91B96a31f0af8DD"
    staticMessageIdMultisigIsmFactory: "0xCb1DC4aF63CFdaa4b9BFF307A8Dd4dC11B197E8f"
    staticAggregationIsmFactory: "0x25CdBD2bf399341F8FEe22eCdB06682AC81fDC37"
    staticAggregationHookFactory: "0xe53275A1FcA119e1c5eeB32E7a72e54835A63936"
    domainRoutingIsmFactory: "0xE2c1756b8825C54638f98425c113b51730cc47f6"
    staticMerkleRootWeightedMultisigIsmFactory: "0x94B9B5bD518109dB400ADC62ab2022D2F0008ff7"
    staticMessageIdWeightedMultisigIsmFactory: "0x70Ac5980099d71F4cb561bbc0fcfEf08AA6279ec"
    proxyAdmin: "0x7e7aD18Adc99b94d4c728fDf13D4dE97B926A0D8"
    mailbox: "0xb1c938F5BA4B3593377F399e12175e8db0C787Ff"
    interchainAccountRouter: "0x4dc4E8bf5D0390C95Af9AFEb1e9c9927c4dB83e7"
    interchainAccountIsm: "0x9F098AE0AC3B7F75F0B3126f471E5F592b47F300"
    validatorAnnounce: "0x79ec7bF05AF122D3782934d4Fb94eE32f0C01c97"
[...]

PR #38, opened 2025-05-05
chore: add celestia-app v4 docker image to fresh compose setup

damiannolan:
## Overview

Adds celestia-app v4 with a fresh compose file for hyperlane/zk-evm testnet. 

The chain state has been initialised with the following and is maintained under `testnet/celestia-validator` for reproducible zero height starts using docker compose:
```
celestia-appd init zkevm-test --chain-id celestia-zkevm-testnet
celestia-appd config set client chain-id celestia-zkevm-testnet
celestia-appd config set client keyring-backend test

# recover the validator key from the hex address in file
celestia-appd keys import-hex validator BAF82CF36B5455C28F8B85DC1F550631884D3643
# add a default user acc key
celestia-appd keys add default 
celestia-appd keys list

celestia-appd genesis add-genesis-account celestia1ex4gt200zvuz3axur6vu3lhnstmzppkcr06zqz 1000000000000utia
celestia-appd genesis add-genesis-account celestia1wy6lkszfvkr7wa9jalvujpd96wa5wep4nljlh8 1000000000000utia
celestia-appd genesis gentx validator 100000000utia --fees 500utia
celestia-appd genesis collect-gentxs
celestia-appd genesis validate
```

Start the celestia-app v4 service via docker-compose:

```bash
docker compose up
```

closes: #31 

The goal is for this to replace the simapp validator and the celestia validator from the existing docker-compose.rollkit.yml file


docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+          "CMD-SHELL",
+          '/bin/sh -c ''BLOCK_HEIGHT=$(curl -sf http://localhost:26657/status | jq -r .result.sync_info.latest_block_height); if [ "$$BLOCK_HEIGHT" -gt 1 ]; then exit 0; else echo "Block height too low: $$BLOCK_HEIGHT"; exit 1; fi''',
+      ]
+      interval: 30s
+      timeout: 5s
+      retries: 3
+      start_period: 10s
+    volumes:
+      - ./testnet/celestia-validator:/setup
+      - ./.tmp/celestia-validator:/home/celestia/.celestia-appd
```
damiannolan, line 20: This uses the same bind mount hack for reproducible chain state which is used in the existing ibc setup.

Is there any reason not to use something like an init-container and named docker volume instead? IMO might be cleaner

e.g. something like this:

```yaml
celestia-init:
    image: ghcr.io/celestiaorg/celestia-app-standalone:02ff9af
    container_name: celestia-init
    entrypoint: sh -c
    command: >
      if [ ! -f /root/.celestia-app/config/genesis.json ]; then
        echo "Initializing Celestia state...";
        celestia-appd init zkevm-test --chain-id celestia-zkevm-testnet &&
        # init keys, gen accs and gen txs...
        celestia-appd genesis collect-gentxs &&
        celestia-appd genesis validate &&
        echo "Init complete.";
      else
        echo "Skipping init, genesis.json already exists.";
      fi
    volumes:
      - celestia-state:/home/celestia/.celestia-app
    restart: "no"
```

Then we do a `depends_on` like:
```yaml
depends_on:
      celestia-init:
        condition: service_completed_successfully
```

damiannolan (commented):
I've left the makefile targets untouched for now. I think its a cleaner approach to start a new compose file and build incrementally rather than touching the existing setup.

Following this PR, the goal would be to:
- start work on https://github.com/celestiaorg/celestia-zkevm-hl-testnet/issues/1 and copy beacond/reth services from existing setup for now
- replace them with official rollkit/evm setup when ready
- incrementally take whats needed and delete whats not needed in this codebase

docker-compose.yml
rootulp: Wow that's cool. IIUC that allows us to 

1. Start + stop + start the celestia-app container and preserve chain history across runs
2. Restart the celestia-app container (maybe with an extra step to `rm genesis.json`) to start a new chain

rootulp (approved):
LGTM, thanks!

docker-compose.yml
damiannolan: Yep, iirc the command you would use to tear down so that the next run would be clean (from zero height) is:
`docker compose down -v` the `-v` flag will remove named volumes.

PR #40, opened 2025-05-06
Replace beacond with Rollkit

Manav-Aggarwal:


testing/files/eth-genesis.json
```diff
@@ -1,358 +1,39 @@
...
-  "nonce": "0x0000000000000000",
+  "nonce": "0x0",
   "timestamp": "0x0",
+  "extraData": "0x",
+  "gasLimit": "0x1c9c38000",
+  "difficulty": "0x0",
+  "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+  "coinbase": "0x0000000000000000000000000000000000000000",
   "alloc": {
-    "0xaF9053bB6c4346381C77C2FeD279B17ABAfCDf4d": {
```
damiannolan, line 33: I think this address has the private key defined in the .env.example file which is useful. 
Would be nice to keep it for testing purposes maybe. 

I'm unsure how I would get the priv keys for the two addresses in the `alloc` block here unless we can add them somewhere in the repo or another env file.

docker-compose.rollkit.yml
```diff
@@ -1,12 +1,17 @@
...
-      - ./testing/files:/beacon-kit/testing/files
-      - ./.tmp:/beacon-kit/.tmp
       - /var/run/docker.sock:/var/run/docker.sock
-    entrypoint: /beacon-kit/testing/files/entrypoint.sh
+    entrypoint: /usr/bin/entrypoint.sh
+    environment:
+      - EVM_ENGINE_URL=http://reth:8551
+      - EVM_ETH_URL=http://reth:8545
+      - EVM_JWT_SECRET=f747494bb0fb338a0d71f5f9fe5b5034c17cc988c229b59fd71e005ee692e9bf
+      - EVM_GENESIS_HASH=0x0a962a0d163416829894c89cb604ae422323bcdf02d7ea08b94d68d3e026a380
```
damiannolan, line 12: Is this the has of the `eth-genesis.json` file?

Manav-Aggarwal: yes, will update it with the changes: https://github.com/rollkit/go-execution-evm?tab=readme-ov-file#reading-genesis-information

testing/files/eth-genesis.json
Manav-Aggarwal: great catch, added it back!

rootulp (approved):
LGTM, thanks!

[follow up] I re-realized that the Rollkit files are in https://github.com/celestiaorg/celestia-zkevm-hl-testnet/tree/main/testing/files and it would be nice to nest them in `testing/files/rollkit` to be consistent with the other directories

Manav-Aggarwal:
> LGTM, thanks!
> 
> [follow up] I re-realized that the Rollkit files are in https://github.com/celestiaorg/celestia-zkevm-hl-testnet/tree/main/testing/files and it would be nice to nest them in `testing/files/rollkit` to be consistent with the other directories



> LGTM, thanks!
> 
> [follow up] I re-realized that the Rollkit files are in https://github.com/celestiaorg/celestia-zkevm-hl-testnet/tree/main/testing/files and it would be nice to nest them in `testing/files/rollkit` to be consistent with the other directories

Follow up issue here: https://github.com/celestiaorg/celestia-zkevm-hl-testnet/issues/44

Issue #39, opened 2025-05-06
Use an init-container for celestia app state in favour of two bind mounts


damiannolan:
This uses the same bind mount hack for reproducible chain state which is used in the existing ibc setup.

Is there any reason not to use something like an init-container and named docker volume instead? IMO might be cleaner

e.g. something like this... if a genesis.json exists then skip init container steps (allows stop/start functionality without compromising state), otherwise a genesis.json doesn't exist - initialise chain state and test keyring.

```yaml
celestia-init:
    image: ghcr.io/celestiaorg/celestia-app-standalone:02ff9af
    container_name: celestia-init
    entrypoint: sh -c
    command: >
      if [ ! -f /root/.celestia-app/config/genesis.json ]; then
        echo "Initializing Celestia state...";
        celestia-appd init zkevm-test --chain-id celestia-zkevm-testnet &&
        # init keys, gen accs and gen txs...
        celestia-appd genesis collect-gentxs &&
        celestia-appd genesis validate &&
        echo "Init complete.";
      else
        echo "Skipping init, genesis.json already exists.";
      fi
    volumes:
      - celestia-state:/home/celestia/.celestia-app
    restart: "no"
```

Then we do a `depends_on` like:
```yaml
depends_on:
      celestia-init:
        condition: service_completed_successfully
```

_Originally posted by @damiannolan in https://github.com/celestiaorg/celestia-zkevm-hl-testnet/pull/38#discussion_r2073289929_

PR #41, opened 2025-05-06
Damian/celestia init container

damiannolan:
## Overview

Introduces an init-container for reproducible chain state while avoiding the bind mount and copy approach, with app home directory under version control.

This allows to stop the docker-compose services with ctrl+c or `docker compose down` while preserving chain state across runs. Restarting with `docker compose up` will continue from the same block height. 

The initialisation logic runs based on the presence of a `genesis.json` file in the docker volume mounted as the app's home directory.

To remove chain state run `docker compose down -v`. Passing the `-v` flag will also removes named docker volumes, allowing you to start from a fresh chain state on the next run. Note, this can be added to documentation later.

closes: #39


docker-compose.yml
```diff
@@ -9,27 +18,24 @@ services:
...
       test: [
         "CMD-SHELL",
-        "BLOCK_HEIGHT=$(curl -sf http://localhost:26657/status | jq -r '.result.sync_info.latest_block_height'); \
-        if [ \"$$BLOCK_HEIGHT\" -gt 1 ]; then \
-          exit 0; \
-        else \
-          echo \"Block height too low: $$BLOCK_HEIGHT\"; \
-          exit 1; \
-        fi"
+        "bh=$(curl -sf http://localhost:26657/status | jq -r '.result.sync_info.latest_block_height'); [ \"$bh\" -gt 1 ]"
```
damiannolan, line 21: drive by update to try make the healthcheck less verbose

scripts/init-celestia.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
+
+GENESIS_FILE="/home/celestia/.celestia-appd/config/genesis.json"
```
rootulp, line 3: [Not blocking] saw a couple PRs to celestia-app to standardize on `./celestia-app` so just flagging that this still uses `./celestia-appd`

```suggestion
GENESIS_FILE="/home/celestia/.celestia-app/config/genesis.json"
```

docker-compose.yml
```diff
@@ -1,4 +1,13 @@
 services:
+  celestia-init:
+    image: ghcr.io/celestiaorg/celestia-app-standalone:02ff9af
+    container_name: celestia-init
+    entrypoint: /scripts/init.sh
+    volumes:
+      - ./scripts/init-celestia.sh:/scripts/init.sh:ro
+      - celestia-app:/home/celestia/.celestia-appd
```
rootulp, line 8: [not blocking] ditto

```suggestion
      - celestia-app:/home/celestia/.celestia-app
```

rootulp (approved):

scripts/init-celestia.sh
damiannolan: 100%, I had made a note of this also. We can rebase the feat branch that Manav made which I took the docker build from and pick up the changes here. 

Let's do that before merging this PR. I can look into it

ninabarbakadze (approved):
nice

scripts/init-celestia.sh
damiannolan: I'll just merge this and create a follow up

PR #42, opened 2025-05-12
chore: add celestia bridge node to compose

damiannolan:
## Overview

- Adds celestia bridge node (v0.23.0-rc0) with v4 support to new docker compose setup.
- Updates celestia-app image to `ghcr.io/celestiaorg/celestia-app-standalone:feature-zk-execution-ism` and removes workarounds for genesis.json

Note, will later extend this setup if a specific account or token is required for the rollup to use the DA layer

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,15 +1,15 @@
 services:
   celestia-init:
-    image: ghcr.io/celestiaorg/celestia-app-standalone:02ff9af
+    image: ghcr.io/celestiaorg/celestia-app-standalone:feature-zk-execution-ism
     container_name: celestia-init
     entrypoint: /scripts/init.sh
     volumes:
       - ./scripts/init-celestia.sh:/scripts/init.sh:ro
-      - celestia-app:/home/celestia/.celestia-appd
+      - celestia-app:/home/celestia/.celestia-app
     restart: "no"
 
   celestia-validator:
-    image: ghcr.io/celestiaorg/celestia-app-standalone:02ff9af
+    image: ghcr.io/celestiaorg/celestia-app-standalone:feature-zk-execution-ism
     container_name: celestia-validator
     ports:
       - "26656:26656"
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -18,13 +18,13 @@ services:
     healthcheck:
       test: [
         "CMD-SHELL",
-        "bh=$(curl -sf http://localhost:26657/status | jq -r '.result.sync_info.latest_block_height'); if [ \"$$bh\" -ge 1 ]; then exit 0; else exit 1; fi"
+        "bh=$(curl -sf http://localhost:26657/status | jq -r '.result.sync_info.latest_block_height'); if [ \"$$bh\" -gt 1 ]; then exit 0; else exit 1; fi"
```
damiannolan, line 21: Quite a subtle one here, but we must use `-gt` rather than `-ge`. 
Using greater than, ensure that we have a committed and produced the first block. This can then report the service as healthy and allow celestia-node to start up successfully. Otherwise, I was running into a `nil commit for block hash`, as the bridge node started prematurely.

docker-compose.yml
```diff
@@ -33,8 +33,26 @@ services:
     networks:
       - celestia-zkevm-net
 
+  celestia-bridge:
+    image: ghcr.io/celestiaorg/celestia-node:v0.23.0-rc0
```
damiannolan, line 37: We can update this with an official release image when its available.

scripts/init-celestia-bridge.sh
damiannolan, line 1: This entrypoint script will check for the presence of an `.env` file and if it exists then source it (as env variables are not persisted across start/stops). Then skip directly to starting the node.

Otherwise, it will reach out to the consensus node `/status` and fetch the genesis block hash, and export that to the `CELESTIA_CUSTOM` var, and save it in the `.env` file. Then proceed with init and starting the node.

rootulp (approved):

Issue #44, opened 2025-05-12
chore: Move Rollkit files into its own directory


Manav-Aggarwal:
LGTM, thanks!
    [follow up] I re-realized that the Rollkit files are in https://github.com/celestiaorg/celestia-zkevm-hl- 
    testnet/tree/main/testing/files and it would be nice to nest them in `testing/files/rollkit` to be consistent with the other 
    directories

_Originally posted by @rootulp in https://github.com/celestiaorg/celestia-zkevm-hl-testnet/pull/40#pullrequestreview-2833919361_

PR #46, opened 2025-05-13
chore: use reth/rollkit images in new compose

damiannolan:
## Overview

This PR takes the `rollkit-evm-single` and `reth` services from #40 and integrates them into the new docker compose file.
I have also addressed the follow up (#44) to maintain files under named service directories under `/testnet`.

closes: #44


damiannolan:
I'll follow this PR with deleting the old compose file, cleaning up unused testing files and ibc related stuff.
I'll also adjust the makefile targets then to use the new compose file, and add some information to the README.md.

docker-compose.yml
```diff
@@ -57,9 +57,63 @@ services:
...
+    entrypoint: /usr/bin/entrypoint.sh
+    environment:
+      - EVM_ENGINE_URL=http://reth:8551
+      - EVM_ETH_URL=http://reth:8545
+      - EVM_JWT_SECRET=f747494bb0fb338a0d71f5f9fe5b5034c17cc988c229b59fd71e005ee692e9bf
+      - EVM_GENESIS_HASH=0x06bab65f0fa3c9b1ba0da34e319d7a5cd948cd10ba33080b88f17d281880e344
+      - EVM_BLOCK_TIME=1s
+      - EVM_SIGNER_PASSPHRASE=secret
+      - DA_ADDRESS=http://celestia-bridge:26658
+      - DA_AUTH_TOKEN=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJBbGxvdyI6WyJwdWJsaWMiLCJyZWFkIiwid3JpdGUiLCJhZG1pbiJdfQ.qxoN_a-k7aZZE77VSpfgyvI_0eNvJ_OgBsT35zxUQnQ
```
damiannolan, line 72: Want to flag here, I have `--rpc.skip-auth` on celestia-node and verified by locally hitting the `node.Ready` jsonrpc. 

Still, without this env variable I got the following, even tho this token doesn't correspond to the account being used in the bridge node.

```
rollkit-evm-single  | 12:36PM ERR RPC call failed error="RPC client error: sendRequest failed: request failed, http status 401 Unauthorized" method=MaxBlobSize module=main
```

docker-compose.yml
```diff
@@ -57,9 +57,63 @@ services:
     networks:
       - celestia-zkevm-net
 
+  rollkit-evm-single:
```
damiannolan, line 60: I removed the unnecessary mounting of the host machine docker socket:

```
volumes:
  - /var/run/docker.sock:/var/run/docker.sock
```

docker-compose.yml
```diff
@@ -57,9 +57,63 @@ services:
...
+    image: ghcr.io/paradigmxyz/reth:v1.3.11
+    container_name: reth
+    ports:
+      - "9001:9001"   # Metrics
+      - "30303:30303" # P2P
+      - "8545:8545"   # HTTP
+      - "8551:8551"   # Auth RPC (Engine API)
+    volumes:
+      - ./testnet/reth:/home/reth
+      - reth:/home/reth/eth-home
```
damiannolan, line 92: Minor tweaks here:
- Bind mount for genesis file and jwt for engine api
- Named docker volume for reth state dir

docker-compose.yml
```diff
@@ -57,9 +57,63 @@ services:
...
+    entrypoint: /bin/sh -c
+    command:
+      - |
+          reth node \
+          --chain /home/reth/eth-genesis.json \
+          --metrics 0.0.0.0:9001 \
```
damiannolan, line 98: Do we need metrics?

docker-compose.yml
```diff
@@ -57,9 +57,63 @@ services:
...
+          --authrpc.addr 0.0.0.0 \
+          --authrpc.port 8551 \
+          --http \
+          --http.addr 0.0.0.0 \
+          --http.port 8545 \
```
damiannolan, line 103: This is default and could be omitted

docker-compose.yml
damiannolan: Any idea why this is the case? cc. @Manav-Aggarwal

docker-compose.yml
rootulp: I don't think we _need_ them b/c we didn't have metrics enabled [here](https://github.com/celestiaorg/celestia-zkevm-ibc-demo/blob/4c43989012340c400d525751c32ef1a2d7762e8f/docker-compose.rollkit.yml#L31-L40) but if this works then seems fine to include

rootulp (approved):

Issue #47, opened 2025-05-13
Cleanup unused files and adjust makefile targets


damiannolan:
> I'll follow this PR with deleting the old compose file, cleaning up unused testing files and ibc related stuff.
I'll also adjust the makefile targets then to use the new compose file, and add some information to the README.md.

_Originally posted by @damiannolan in https://github.com/celestiaorg/celestia-zkevm-hl-testnet/issues/46#issuecomment-2876640449_

Issue #33, opened 2025-05-01
Modify `make transfer` to use Hyperlane Multisig ISM


Manav-Aggarwal:
Currently the `make transfer` command triggers a transfer from simapp to the EVM rollup via the established IBC clients. We should modify the command to trigger a transfer from Celestia to the EVM rollup via the established Hyperlane Warp Route which should be a Multisig ISM

PR #48, opened 2025-05-13
chore: add initial hyperlane configs for e2e transfer test

damiannolan:
## Overview

NOTE: The hyperlane CLI is currently only able to support EVM chains. This PR adds a Go program to facilitate deployment of basic components on the cosmosnative module for testing.

This PR adds:
- A complete `README.md` for successful e2e warp transfer from Celestia to EVM (Reth)
- A local Hyperlane registry and deployment configs for EVM via Hyperlane CLI
- A golang CLI for deployment of Hyperlane components on cosmosnative (celestia)
- Relayer configuration for use with the provided deployment configs

Edit: I'll adjust the Makefile targets in a follow up PR and then we can close out a few of the issues. #1 #47 #33 come to mind


damiannolan:
I have already started work on a Dockerfile setup to automate the deployment process for both EVM contracts and cosmosnative components. 

The goal is to streamline the deployment steps and integrate the docker image for the hyperlane relayer into the compose services using the config in this PR.

hyperlane/registry/strategies/default-strategy.yaml
damiannolan, line 1: From my investigation I found that the hyperlane CLI requiresa "strategy" file for deploying on cosmosnative modules which will look like this. This is also not being used rn as the support isn't fully ready yet.

hyperlane/relayer-config.json
damiannolan, line 1: I had to write this by hand modelled off https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/rust/main/config/testnet_config.json

hyperlane/configs/warp-config.yaml
damiannolan, line 1: My understanding is that when there's cosmosnative support in the CLI, then this file would be extended with a collateral token definition for utia

PR #49, opened 2025-05-13
chore: delete simapp, ibc and legacy test network files

damiannolan:
## Overview

This PR deletes unused files:
- simapp
- ibc client and x/header modules
- testnet files for ibc-zkevm demo


damiannolan:
There is more to delete but this should be fine for this PR. I can follow up with another.
IMO it makes more sense to take whats needed from the zkevm-ibc demo when needed and treat this a fresh.

rootulp (approved):

PR #51, opened 2025-05-19
feat: add dockerfile and entrypoint script to deploy hyperlane

damiannolan:
## Overview

- Adds a dockerfile to automate the deployment of hyperlane components on EVM and cosmosnative (celestia).
- Adjusts the Makefile targets: `start, stop` to use the new docker-compose services.

closes: #1


.github/workflows/docker.yml
damiannolan, line 1: simple build and push workflow, if we need to adjust this to use the one provided by celestia ops team then we can do that later, afaik the only difference is certain arch support for running k8s which we don't need here

damiannolan (commented):
Ideally would like to keep PRs well scoped so I will publish this for review and create more follow up PRs.

Next I would like to:
- Clean out unused files from the ibc demo which are bloating this repo #47 
- Address the comment I left about extending the init container for hyperlane deployment to complete the last step. #53 
- Add the hyperlane relayer to docker compose #54

hyperlane/Dockerfile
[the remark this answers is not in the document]
damiannolan: https://github.com/celestiaorg/celestia-zkevm-hl-testnet/issues/53

.github/workflows/docker.yml
damiannolan: This _should_ just work when merged to main. Can take a look when this PR lands

rootulp (approved):
LGTM

.github/workflows/docker.yml
damiannolan: lol

Issue #54, opened 2025-05-20
Add the hyperlane relayer to the docker compose services


damiannolan:
See: https://github.com/celestiaorg/celestia-zkevm-hl-testnet/pull/51#pullrequestreview-2854290029

Add the hyperlane lane relayer to the docker-compose services. Mount the config file provided in `hyperlane/relayer-config.json`.

Image: gcr.io/abacus-labs-dev/hyperlane-agent:agents-v1.4.0

Issue #53, opened 2025-05-20
Extend hyperlane-init image to install foundry tools and complete remote router registration


damiannolan:
> This should be extended to install foundry tools in order to complete the last step: `enroll-remote-router` on both celestia-app and evm.

> I think this can be done in a follow up PR.

_Originally posted by @damiannolan in https://github.com/celestiaorg/celestia-zkevm-hl-testnet/pull/51#discussion_r2098034017_

PR #56, opened 2025-05-21
feat: complete setup and add relayer to docker compose

damiannolan:
## Overview

- Finishes the automation of hyperlane provisioning.
- Adds whats needed to run the relayer service in docker compose (but commented out for now).

closes: #53 

partially-closes: #54 (relayer is added but seems like the container version doesn't work the same as the binary I built and tested from main locally)


hyperlane/cmd/hyp/cmd/cmd.go
```diff
@@ -123,3 +125,51 @@ func getDeployCmd() *cobra.Command {
...
+				Owner:   broadcaster.address.String(),
+				TokenId: tokenID,
+				RemoteRouter: &warptypes.RemoteRouter{
+					ReceiverDomain:   1234,
+					ReceiverContract: args[3],
+					Gas:              math.ZeroInt(),
+				},
+			}
+
+			res := broadcaster.BroadcastTx(ctx, &msgEnrollRemoteRouter)
```
damiannolan, line 168: I ran into an issue with this because we're forced to use sign mode amino json here.

```
sign tx failed: hyperlane.warp.v1.RemoteRouter: {TagNum: 1, WireType:"varint"}: unknown protobuf field [/go/pkg/mod/cosmossdk.io/errors@v1.0.2/errors.go:155]
```

Essentially the sign mode handler calls into `doRejectUnknownFields()` [ref](https://github.com/celestiaorg/cosmos-sdk/blob/release/v0.50.x-celestia/codec/unknownproto/unknown_fields.go#L47) in the depths of the sdk. It then fails to find a fulfilled msg descriptor for the type:

image

I can't further debug this. The go pkg is being imported and therefore, the `init()` func is called by the package loader which _should_ register the protobuf types in the global registry. 
Why its not being found is beyond me. I've faced errors like this before with the sdk and they are a nightmare to debug. 

Furthermore, there seems to be a bug with using sign mode direct in the sdk as we hit:
```
raw_log: 'signature verification failed; please verify account number (10) and chain-id
  (celestia-zkevm-testnet): (unable to verify single signer signature): unauthorized'
```

I think the best opt here is to just use the `celestia-appd` binary directly.

docker-compose.yml
```diff
@@ -125,6 +125,22 @@ services:
     networks:
       - celestia-zkevm-net
 
+  # TODO: Determine why this does not work but running a relayer binary locally built from main
```
damiannolan, line 128: I reached out to the hyperlane team about this

rootulp (approved):

PR #57, opened 2025-05-21
chore: add temporary local relayer binary and make target

damiannolan:
## Overview

We can temporarily rely on this local relayer to get the job done. 
I've checked in a binary I built from hyperlane-monorepo@main last week. 

The binary does appear to be a little bit large (~81mb).
If we'd prefer to just provide a commit and instructions to compile the binary we can do that. I am indifferent.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ build
 /target
 *.bin
 bin
+
+**/hyperlane_db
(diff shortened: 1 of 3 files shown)


rootulp (approved):
I'm fine w/ the binary checked in 👍

PR #62, opened 2025-06-19
chore: add evm-prover service skeleton

damiannolan:
## Overview

Adds a skeleton evm-prover grpc service to consume sp1 programs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1208,18 +1208,46 @@ checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
 
 [[package]]
 name = "axum"
+version = "0.6.20"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3b829e4e32b91e643de6eafe82b1d90675f5874230191a4ffbc1b336dec4d6bf"
+dependencies = [
+ "async-trait",
+ "axum-core 0.3.4",
+ "bitflags 1.3.2",
+ "bytes",
+ "futures-util",
+ "http 0.2.12",
+ "http-body 0.4.6",
+ "hyper 0.14.32",
+ "itoa",
+ "matchit",
+ "memchr",
+ "mime",
+ "percent-encoding",
+ "pin-project-lite",
+ "rustversion",
+ "serde",
+ "sync_wrapper 0.1.2",
+ "tower 0.4.13",
+ "tower-layer",
+ "tower-service",
+]
+
+[[package]]
+name = "axum"
 version = "0.7.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "edca88bc138befd0323b20752846e6587272d3b03b0343c8ea28a6f819e6e71f"
 dependencies = [
  "async-trait",
- "axum-core",
+ "axum-core 0.4.5",
  "bytes",
  "futures-util",
  "http 1.3.1",
  "http-body 1.0.1",
  "http-body-util",
- "hyper",
+ "hyper 1.6.0",
  "hyper-util",
  "itoa",
  "matchit",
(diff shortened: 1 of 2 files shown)


crates/evm-prover/build.rs
```diff
@@ -0,0 +1,31 @@
...
+        .build_server(true)
+        .build_client(true)
+        .file_descriptor_set_path("src/proto/descriptor.bin")
+        .out_dir("src/proto")
+        .compile(&proto_files, &["../../proto"])?;
+
+    // This build.rs can be extended to compile the SP1 programs and output the elfs to a standard path
+    // from which we can then consume them: e.g. output to evm-prover/artifacts or evm-prover/elf
+    // consume from prover/prover.rs with include_elf!("../../artifacts/evm-exec-elf")
+    // ref: https://docs.succinct.xyz/docs/sp1/writing-programs/compiling#build-script
```
damiannolan, line 28: will add sp1 program build directives here later

crates/evm-prover/src/config/config.rs
```diff
@@ -0,0 +1,17 @@
+use serde::{Deserialize, Serialize};
+
+pub const APP_HOME_DIR: &str = ".evm-prover";
+pub const CONFIG_FILE: &str = "config.yaml";
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Config {
+    pub grpc_address: String,
+}
```
damiannolan, line 9: this can be easily extended with other things we will eventually need

crates/evm-prover/src/prover/prover.rs
```diff
@@ -0,0 +1,3 @@
+//! TODO: Types for block prover will live here
+//!
+//! pub struct BlockProver {}
```
damiannolan, line 3: The plan is to put a `BlockProver` struct here which will encapsulate interactions with sp1 ProverClient and rpcs for data inputs. Basically refactor out a lot of the blevm stuff

damiannolan (commented):
Half of the diff here is proto gen. Don't be afraid 🙏🏻 😅

rootulp (approved):
[follow up] it may be helpful to include docs that describe how maintainers can regenerate the proto-generated files (i.e. `celestia.prover.v1.rs`) b/c I didn't find a Makefile command for it. 

So either add a Makefile command (preferable) or a command in README

Ref: https://github.com/celestiaorg/celestia-zkevm-hl-testnet/blob/main/Makefile

damiannolan:
> [follow up] it may be helpful to include docs that describe how maintainers can regenerate the proto-generated files (i.e. `celestia.prover.v1.rs`) b/c I didn't find a Makefile command for it.
> 
> So either add a Makefile command (preferable) or a command in README
> 
> Ref: https://github.com/celestiaorg/celestia-zkevm-hl-testnet/blob/main/Makefile

Aye, good attention to detail! I can add some docs to the README of the prover service. 
The protobufs are generated by the custom `build.rs` which lives alongside the `src` directory in `evm-prover`. So if there is any changes they will be built when running `cargo build`.

It includes these lines which walks the proto dir and collects the files, then builds them outputting the `descriptor.bin`(which we can probably actually gitignore):
```rust
    // Collect all .proto files recursively under /proto/
    let proto_files: Vec<PathBuf> = WalkDir::new("../../proto")
        .into_iter()
        .filter_map(|entry| {
            let path = entry.ok()?.path().to_path_buf();
            if path.extension().map_or(false, |ext| ext == "proto") {
                Some(path)
            } else {
                None
            }
        })
        .collect();

    tonic_build::configure()
        .build_server(true)
        .build_client(true)
        .file_descriptor_set_path("src/proto/descriptor.bin")
        .out_dir("src/proto")
        .compile(&proto_files, &["../../proto"])?;
```

I also consumed that `descriptor.bin` file in order to register the reflection service, similarly to how we normally have in Go grpc services. This let's us use helpful tools like `grpcurl` etc

rootulp:
Oops sorry I totally missed that in the review. Auto-generating the protobufs sounds great 👍

PR #69, opened 2025-06-24
chore: add parser script, remove custom serde and regen testdata

damiannolan:
## Overview

This PR aligns with the ISM testdata PR from celestia-app here: https://github.com/celestiaorg/celestia-app/pull/5041

- Adds a parser script to write groth16 proof and public values as separate artifacts to allow for easier testing in Go
- Removes the custom `hex_bytes` serde in favour of `fmt::Display` trait impls
  - This reduces cycles, and also encodes the public values correctly so that they can be parsed in Go (context added in comments)
- Regenerates testdata for a range of blocks where state root changes (so that testdata is more meaningful)


damiannolan (commented):
add context on custom serde removal

Issue #65, opened 2025-06-20
Implement the BlockProver struct in evm-prover


damiannolan:
The plan is to put a `BlockProver` struct here which will encapsulate interactions with sp1 ProverClient and rpcs for data inputs. Basically refactor out a lot of the blevm stuff

_Originally posted by @damiannolan in https://github.com/celestiaorg/celestia-zkevm-hl-testnet/pull/62#discussion_r2157312701_

PR #72, opened 2025-06-26
feat: add program prover trait and skeleton implementation

damiannolan:
## Overview

Part of ref: #65 

This PR adds new trait `ProgramProver` to be implemented per sp1 program within a prover service. 
The trait defines the base functionality which should be implemented for a `ProgramProver`, allowing generic plugging of its `Input` and `Output` structs. This allows us to have separation of concerns with respect to how the prover service interacts with an SP1 program, i.e. not mixing logic such as rpc data retrieval, etc which should be self contained elsewhere.

### DIFF
--- a/crates/evm-prover/src/prover/prover.rs
+++ b/crates/evm-prover/src/prover/prover.rs
@@ -1,3 +1,16 @@
+use std::sync::Arc;
+
+use alloy_provider::ProviderBuilder;
+use anyhow::Result;
+use celestia_rpc::{BlobClient, Client, HeaderClient, ShareClient};
+use celestia_types::nmt::Namespace;
+use celestia_types::{Blob, Commitment, ExtendedHeader, ShareProof};
+use eq_common::KeccakInclusionToDataRootProofInput;
+use reth_chainspec::ChainSpec;
+use rsp_host_executor::EthHostExecutor;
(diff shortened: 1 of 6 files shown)


crates/evm-prover/src/prover/prover.rs
```diff
@@ -1,9 +1,119 @@
...
+        let output = BlockExecOutput {};
+        output
+    }
+
+    fn prover(&self) -> &EnvProver {
+        &self.prover
+    }
+}
+
+impl BlockExecProver {
```
damiannolan, line 97: lower level details such as data retrieval from rpc etc can go here, as they're not as tightly coupled with program proving

crates/evm-prover/src/prover/prover.rs
```diff
@@ -1,9 +1,119 @@
...
+        let output = self.post_process(proof.clone());
+
+        Ok((proof, output))
+    }
+
+    /// Returns the SP1 Prover.
+    fn prover(&self) -> &EnvProver;
+
+    /// Parse or convert program outputs.
+    fn post_process(&self, proof: SP1ProofWithPublicValues) -> Self::Output;
```
damiannolan, line 60: its possible we don't really need this and it can be removed.

Issue #50, opened 2025-05-14
Rollkit docker-compose configuration doesn't allow start/stop


damiannolan:
## Overview

Stopping the `docker compose` set of services and restarting without tearing down volumes leads to the `rollkit-evm-single` container failing:

```
2:56PM INF successfully submitted Rollkit headers to DA layer daHeight=0 gasPrice=-1 headerCount=6 module=BlockManager
2:56PM INF Creating empty block height=513 module=BlockManager
2:56PM INF Creating empty block height=514 module=BlockManager
2:56PM INF Creating empty block height=515 module=BlockManager
2:56PM INF Creating empty block height=516 module=BlockManager
2:56PM INF Creating empty block height=517 module=BlockManager
2:56PM INF Creating empty block height=518 module=BlockManager
2:56PM INF Creating empty block height=519 module=BlockManager
2:56PM INF Creating empty block height=520 module=BlockManager
2:56PM INF Creating empty block height=521 module=BlockManager
2:56PM INF successfully submitted Rollkit headers to DA layer daHeight=0 gasPrice=-1 headerCount=6 module=BlockManager
Error: failed to initialize signer: key file already exists at /root/.evm-single/config/signer.json
Usage:
  evm-single init [flags]

Flags:
      --chain_id string                                         chain ID (default "rollkit-test")
  -h, --help                                                    help for init
      --rollkit.da.address string                               DA address (host:port) (default "http://localhost:7980")
```

See logs. When services are restarted, the node tries to reinitialise and fails.

## Proposal

Override the entrypoint with a custom `docker-entrypoint.sh` script.
Check for the presence of the `signer.json`, if it doesn't exist then run the existing entrypoint under `/usr/bin/entrypoint.sh`, otherwise just start the sequencer so it can continue gracefully.

PR #75, opened 2025-06-28
fix: update rollkit containers and allow rollkit restart without tearing down volumes

Manav-Aggarwal:
## Overview

Should allow rollkit-evm-single service to restart cleanly.

As a user of the docker compose services I would like to stop the services running on my machine and restart them while preserving the existing chain state. Right now we are blocked on achieving this only by the rollkit-evm-single service.

closes: https://github.com/celestiaorg/celestia-zkevm-hl-testnet/issues/50

Follow up issue: https://github.com/celestiaorg/celestia-zkevm-hl-testnet/issues/76

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -98,6 +98,7 @@ services:
       - |
           lumen node \
             --chain /home/reth/eth-genesis.json \
+            --datadir /home/reth/eth-home \
             --metrics 0.0.0.0:9001 \
             --log.file.directory /home/reth/logs \
             --authrpc.addr 0.0.0.0 \
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -58,14 +58,16 @@ services:
       - celestia-zkevm-net
 
   rollkit-evm-single:
-    image: ghcr.io/rollkit/rollkit-evm-single:v0.1.3
+    image: ghcr.io/rollkit/rollkit-evm-single:pr-2394
     container_name: rollkit-evm-single
     entrypoint: /usr/bin/entrypoint.sh
+    volumes:
+      - evm-single-data:/data
     environment:
       - EVM_ENGINE_URL=http://reth:8551
       - EVM_ETH_URL=http://reth:8545
       - EVM_JWT_SECRET=f747494bb0fb338a0d71f5f9fe5b5034c17cc988c229b59fd71e005ee692e9bf
-      - EVM_GENESIS_HASH=0x06bab65f0fa3c9b1ba0da34e319d7a5cd948cd10ba33080b88f17d281880e344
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -58,14 +58,16 @@ services:
       - celestia-zkevm-net
 
   rollkit-evm-single:
-    image: ghcr.io/rollkit/rollkit-evm-single:v0.1.3
+    image: ghcr.io/rollkit/rollkit-evm-single:pr-2394
     container_name: rollkit-evm-single
     entrypoint: /usr/bin/entrypoint.sh
+    volumes:
+      - evm-single-data:/data
```
damiannolan, line 65: is the rollkit/sequencer block store maintained under a directory called `/data` at the root of the filesystem?

would be nice to put this into some standard home directory eventually, but it can be done later, would've execpted it at somewhere like `~/.rollkit/data`

docker-compose.yml
```diff
@@ -80,33 +82,43 @@ services:
...
-          --chain /home/reth/eth-genesis.json \
-          --metrics 0.0.0.0:9001 \
-          --authrpc.addr 0.0.0.0 \
-          --authrpc.port 8551 \
-          --http \
-          --http.addr 0.0.0.0 \
-          --http.port 8545 \
-          --http.api "eth,net,web3,txpool" \
-          --authrpc.jwtsecret /home/reth/jwt.hex \
-          --datadir /home/reth/eth-home \
```
damiannolan, line 106: How come this was removed?

I believe this is required in order to maintain the chain db between restarts. The rollkit restart seems to be fixed here, but when this is missing, the volume data expected to be under `/home/reth/eth-home` is then missing on restart, and all contract deployments etc are lost.

For example, when I restart and attempt to query my hyperlane synthetic ERC20 I get:
```shell
cast code 0xa7578551baE89a96C3365b93493AD2D4EBcbAe97
0x
```

damiannolan (commented):
Nice work, thank you @Manav-Aggarwal! 

Left a couple of comments. I think we can merge this once we fix the reth volume issue

docker-compose.yml
```diff
@@ -80,33 +82,43 @@ services:
...
-          --ipcpath /home/reth/eth-home/eth-engine.ipc \
-          --disable-discovery \
-          --rpc.eth-proof-window 120000
+          lumen node \
+            --chain /home/reth/eth-genesis.json \
+            --metrics 0.0.0.0:9001 \
+            --log.file.directory /home/reth/logs \
```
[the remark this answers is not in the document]
damiannolan: https://github.com/celestiaorg/celestia-zkevm-hl-testnet/pull/75/commits/351f88f432353e278d75ea84329fcf257f9e1547

PR #77, opened 2025-06-30
chore: expose rollkit rpc for docker network

damiannolan:
## Overview

Exposes rollkit's rpc listen address. Default host: 127.0.0.1 needs to be: 0.0.0.0 host
Requires copying the entrypoint script as its not easily extendable. I opened an issue in rollkit.

With this I can query store metadata to get the da included height:

```
curl -X POST \
  -H "Content-Type: application/json" \
  -d '{"key": "rhb/100/h"}' \
  http://localhost:7331/rollkit.v1.StoreService/GetMetadata
{"value":"HgAAAAAAAAA="}                                                                              
```

This value seems to be base64 encoded little endian bytes of the uint64: 30

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -60,8 +60,11 @@ services:
   rollkit-evm-single:
     image: ghcr.io/rollkit/rollkit-evm-single:pr-2394
     container_name: rollkit-evm-single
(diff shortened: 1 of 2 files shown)


damiannolan:
This is required to progress #73

rootulp (approved):

Issue #85, opened 2025-07-03
Program design refactor for EVM execution and data availability


damiannolan:
Currently the SP1 program [`evm-exec`](https://github.com/celestiaorg/celestia-zkevm-hl-testnet/tree/main/crates/sp1/evm-exec) proves single EVM block execution (STF) and its blob data availability within celestia. 

This issue proposes a redesign to instead, prove `n` data blobs and their associated EVM STFs within a single celestia block, essentially inverting the design. 
This serves to support a design where a celestia block may have zero or many EVM blocks posted to the namespace within it.

This is a placeholder and should be fleshed out as the discussion progresses and the design becomes more mature.

cc. @cmwaters

Issue #83, opened 2025-07-03
Refactor the evm-exec datagen script to use new rollkit metadata in favour of indexer


damiannolan:
The `datagen.rs` script in evm-exec-script used for crawling the compose network for proof input data and writing to files relies on the old ibc-demo indexer service - which admittedly, I the culprit, was running locally to pull in proof data.

Now that we have updated to the newer rollkit and reth dependencies, we can refactor this script to use the `GetMetadatRequest` rpcs from rollkit in order to retrieve the appropriate inclusion height for celestia.

With this issue, the `datagen` script will be reliable and reusable against the docker compose deployment in this repo.

Issue #79, opened 2025-07-03
Data is no longer posted for empty EVM blocks


damiannolan:
## Overview 

Since upgrading the rollkit and reth software in #75, there has been a number of changes.

- The `Data` type previously posted by rollkit has been changed. Rollkit now posted `SignedData` for its tx payloads to Celestia. 
- `SignedData` PFBs are not posted for empty EVM blocks.

This needs to be handled both in the SP1 program (`evm-exec`) as well as the `evm-prover` service which fetches the input data. 

Note, #78 introduces data fetching for the `BlockExecProver`, but should be refactored to handle this in a follow up PR since it is effectively a breaking change.

This requires some further discussion on how the SP1 program should handle this with respect to proving equivalency between tx data blobs and STF output headers (i.e. #68)

Issue #37, opened 2025-05-04
Modify proof system to only require the last celestia header hash


cmwaters:
Currently, we require that the public aggregated input provide all celestia headers across the range of data that the EVM rollup is trying to prove across. This requires Celestia to persist all these headers (hence the x/headers module).

https://github.com/celestiaorg/celestia-zkevm-hl-testnet/blob/477b43bd181041a2a2b8ab224793d5576404bfa2/provers/blevm/common/src/lib.rs#L26

More optimal would be to have those headers passed through in the private input and merely assert that the last header matches the one provided in the public input thus the public input only requires the last header hash (which is provided in the context and thus doesn't require the x/headers module)

cmwaters:
It's feasible that we don't need the Celestia header hashes at all if we can modify Rollkit such that Celestia headers are embedded in Rollkit state root

Thus the prover program would take the trusted celestia state within the trusted rollup state and use conventional header sync up to the header hash in the final state. This would lead to a proof design that could also be used to bridge to non Celestia chains

damiannolan:
Can you take a look at https://github.com/celestiaorg/celestia-zkevm-hl-testnet/blob/main/crates/sp1/evm-range-exec/program/src/main.rs#L62-L89 and see if this satisfies the initial point?

The assumption here is that rollkit blobs are posted either in the same celestia block or in the a sequential celestia block.

> thus the public input only requires the last header hash (which is provided in the context and thus doesn't require the x/headers module)

I don't think we can confidently say that the celestia `ctx.Header()` will be the latest header input to the circuit at the time the tx is processed on celestia. We would need the latest celestia header hash which corresponds to the latest inclusion height of the rollkit/evm block we are updating to.

By the time the proof is produced for N evm blocks included at latest celestia height H, celestia will already have produced H+x blocks

PR #87, opened 2025-07-04
refactor: sp1 evm block execution and data availability program redesign

damiannolan:
## Overview

This PR refactors the `evm-exec` program design. 
Rather than proving a single EVM block and its data availability in Celestia, we now prove all EVM blocks within a single Celestia block. This allows us to commit to the previous Celestia header hash and current Celestia header hash via the Header provided as input, and in-turn, to verify sequentiality of Celestia headers within the range circuit.
As well as this, it allows us to add support for proving zero, one or more EVM blocks within a Celestia block using the range circuit.

This PR also removes the `testdata/inputs` pulled from the previous rollkit/reth images and adds new data inputs.
Those new inputs are used in #88 where new proofs have been added to the testdata directory.

closes: #37 
closes: #79
closes: #85
closes: #83


crates/sp1/evm-exec/script/src/bin/main.rs
```diff
@@ -103,24 +104,70 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-        bincode::deserialize(&fs::read(format!("{input_dir}/data_root_proof.bin"))?)?;
-    stdin.write(&data_root_proof);
+    let share_proofs = read_share_proofs(input_dir)?;
+    stdin.write(&share_proofs);
 
     Ok(())
 }
+
+/// Reads and deserializes ordered `EthClientExecutorInput` files from the given directory.
+fn read_client_inputs(input_dir: &str) -> Result<Vec<EthClientExecutorInput>> {
```
damiannolan, line 123: both `read_client_inputs` and `read_share_proofs` fns are pretty LLM lol.. 

they basically just crawl the input directory for files which match regexes like this and return vectors of them:
- `client_input-{block-number}.bin`
- `share_proof-{block-number}.bin`

crates/sp1/evm-range-exec/program/src/main.rs
```diff
@@ -62,36 +57,40 @@ pub fn main() {
...
-            curr.prev_header_hash
+            prev.new_height,
+            curr.prev_height
         );
 
-        // Only check sequentiality if the inclusion height changed.
-        if curr.celestia_header_hash != prev.celestia_header_hash {
-            assert_eq!(
-                curr.prev_celestia_header_hash,
-                prev.celestia_header_hash,
```
damiannolan, line 82: Previously this meant that if a rollup missed a celestia block height and failed to include something that the assumptions of the proof system would be broken.

The new approach allows support for zero, one or more rollup blocks in the celestia block

crates/sp1/evm-exec/script/src/bin/datagen.rs
```diff
@@ -1,61 +1,50 @@
 //! A script that generates proof input data for a range of EVM blocks.
```
damiannolan, line 1: Most diffs are probably in this file which just scrapes docker compose services for proof inputs.

This previously relied on a hacked indexer from the ibc-demo that I used first time to get proof inputs. That's no longer the case now and uses metadata requests to the rollkit sequencer to get inclusion heights for data.

rootulp (approved):
LGTM! Really great work (as usual)

Issue #68, opened 2025-06-23
Verify blob proof data is equal to the STF data


damiannolan:
## Overview 

Right now we accept a blob proof and an EVM stf (rsp client executor input).
We need to validate that the blob which we are verifying in celestia is the the same data which we are executing in the stf.

This is essentially validating equivalency between the state transition function execution data and the data which has been posted to celestia data availability layer.

damiannolan:
I have been trying to look into this a bit, and right now having difficultly understanding exactly the data which is encoded into the blob and its format. 

There is discrepancies between what is available on rollkit/main branch and some historical code which is in the archived go-execution-evm repository. 

The code on main seems to be encoding `types.Transaction` from `"github.com/ethereum/go-ethereum/core/types"`.
Whereas the legacy code follows a different approach, whereby the first tx is json encoded as a `engine.ExecutionPayloadEnvelope` and txs thereafter are slotted into this(afaict). 

It's unclear to me what approach is being used in `rollkit-evm-single:v0.1.3` which is currently used in this repo. I've tried unmarshalling the first tx from rollkit's `Data` type [ref](https://github.com/rollkit/rollkit/blob/main/proto/rollkit/v1/rollkit.proto#L95) as both of the above and both fail.

For example:

```golang
tx := new(types.Transaction)
if err := tx.UnmarshalBinary(rawTx); err != nil {
	log.Printf("error unmarshalling tx binary: %v", err)
	return 0, err
}
```

and

```golang
var payload engine.ExecutionPayloadEnvelope
if err := json.Unmarshal(rawTx, &payload); err != nil {
	fmt.Printf("failed to unmarshal paylod: %v\n", err)
}
```

We should be able to validate the blob contents against the `client_executor_input.current_block` [ref](https://github.com/succinctlabs/rsp/blob/main/crates/executor/client/src/io.rs#L35-L39)

damiannolan:
I have a branch for this feature here: https://github.com/celestiaorg/celestia-zkevm-hl-testnet/blob/damian/blob-equivalency-verification/crates/sp1/evm-exec/program/src/main.rs#L124-L140 

I just need to test it properly with the new input data, as the testdata we currently have is not using `SignedData`.
Hopefully I will have that today

PR #88, opened 2025-07-07
feat: add blob equivalency verification

damiannolan:
## Overview

This PR adds blob data equivalency checks against the STF output.

closes: #68


crates/sp1/evm-exec/program/src/main.rs
```diff
@@ -87,8 +94,31 @@ pub fn main() {
             header.number
         ));
 
-        // TODO: Verify blob tx data equivalency against header.transactions_root
-        // https://github.com/celestiaorg/celestia-zkevm-hl-testnet/issues/68
+        let shares: Vec<Share> = blob_proof
+            .shares()
+            .iter()
+            .map(|b| Share::from_raw(b).expect("Failed to parse raw shares"))
+            .collect();
```
cmwaters, line 101: I didn't realise that the blob_proof contained all the raw shares? Does that mean we are submitting the transactions twice: once in ClientExecutorInput and the other in each of the blob proofs? 

I thought what we would do is the other way around: Turn the transaction in the client executor input into blobs and assert that the blobs, given the blob proof, were part of the data root

cmwaters (commented):
It looks like you are asserting that the raw shares

PR #91, opened 2025-07-09
chore: update rollkit crate deps to upstream

damiannolan:
## Overview

Updates the `rollkit-types` crate to the upsteam repository and regnerates the proofs due to the dep change.


Cargo.toml
```diff
@@ -15,19 +15,22 @@ alloy-provider = { version = "0.12.6" }
...
 nmt-rs = "*"
+prost = "0.12"
 reqwest = "0.12.20"
 reth-chainspec = { git = "https://github.com/paradigmxyz/reth", tag = "v1.3.3", default-features = false }
 reth-primitives = { git = "https://github.com/sp1-patches/reth", tag = "rsp-20240830", default-features = false, features = [
     "alloy-compat",
     "optimism",
     "std",
 ] }
+rollkit-types = { git = "https://github.com/rollkit/rollkit", branch = "main", package = "rollkit-types", default-features = false }
```
rootulp, line 33: LGTM. My only concern is that it seems like recent Rollkit changes have caused breaks in this repo. Using `main` here may also cause breaks in the future.

I wonder if you think it's safer to pin to a commit (i.e. latest commit on main) so that more changes in RollKit don't cause churn in this repo.

rootulp (approved):

Cargo.toml
damiannolan: I can't argue with that! :D 

I'll lock in on a rev and we'll leave it there until they've pushed to crates.io!

Issue #89, opened 2025-07-08
Proof construction should be deterministic (and not allow equivocation)


cmwaters:
One of the properties of correctness for our proof system is:

- Non equivocation: A prover should not be capable, given the same "previous rollup state" and "celestia data root", to produce two different but valid "new rollup state"'s.

<img width="401" height="161" alt="Image" src="https://github.com/user-attachments/assets/361c1cd0-63d9-4ca2-a566-3e6e0d105b8b" />
 
In the current circuit design, we take merkle proofs that prove the inclusion of PFBs to the data root, we validate that the `SignedData` was signed by the sequencer and that they are ordered by the height of the `SignedData` (or by proxy, the `SignedHeader`). This would allow a malicious prover, to generate two proofs: one using blobs 10 and 11a and the other using blobs 10 and 11b. Both would lead to a valid proof that give verifiers different final states.

## Solutions

There are two approaches:
- Read all (or at least the first n) blobs in a namespace: By reading and proving all, the circuit can use the ordering of consensus to know whether 11a or 11b is the valid path. Using first come first served, only 11a would be valid for all verifiers of the proof. This has the downside of namespace exhaustion.
- Read the sequence number of the PFB. Each PFB is wrapped by a transaction. The transaction has a monotonically increasing sequence number. By reading this and asserting that 11a comes directly after 10. The circuit will not allow a construction that uses 11b as it has a sequence two greater than the sequence of the tx at rollup height 10. This approach, however, imposes a few constraints: Only one blob per tx and the sequencer keys can not be used to sign anything else else it would increment the sequence number and break the rollup.

damiannolan:
So maybe in simpler terms, this is essentially that the proof system requires double signing protection, such that a prover given inputs 11a OR 11b would produce two valid proofs with different outputs, correct?

adlerjohn:
For (1) you'd only need to read the first part of each blob (like a header), no?

damiannolan:
I think if some approach similar to above was taken, then we would also need to prove the bounds, i.e. any shares less than `start` and greater than `end - 1` live in a different namespace.

damiannolan:
We should also be able to prove completeness of the namespace by accepting the `ExtendedHeader` rather than just the tendermint header to the program. We can then query [`GetNamespaceData`](https://node-rpc-docs.celestia.org/?version=v0.22.3#share.GetNamespaceData) which will return `n` NMTProofs for each row containing our namespace data.

[...]

Issue #94, opened 2025-07-15
Avoid executing `EthClientExecutorInput` for empty blocks and support empty celestia blocks


damiannolan:
## Overview

The evm-exec program currently expects `n` evm state transition functions via `Vec<EthClientExecutorInput>` for each and every evm block posted to celestia. [ref](https://github.com/celestiaorg/celestia-zkevm-hl-testnet/blob/main/crates/sp1/evm-exec/program/src/main.rs#L46)

This means that even for empty EVM blocks, we are executing a `EthClientExecutorInput` which is costly on cycles yet it yields no change to our application state root. 

The list of executor inputs is currently based on the number of `SignedHeader` blobs included in the celestia block.
We rely on the first and last entries of the list of executor inputs to commit our outputs: 
- New height
- New state root
- Previous height
- Previous state root

This poses a problem whenever a celestia block does not include any data for an EVM block, as we will have no handle on where the EVM application state is at that given time. 
The proof system should be able to support a celestia block which contains zero rollup/EVM app data blobs.
Furthermore, the proof system should only be interested when there _is_ tx data to execute and prove inclusion in the celestia namespace.

## Proposal

Take `TrustedStateRoot` and `TrustedHeight` as inputs to the SP1 program (independent of EthClientExecutorInputs).
In the event of empty blob data in the celestia block we commit the trusted values as both previous and new values (height and state root).

By taking the `TrustedStateRoot` and `TrustedHeight` independently of `EthClientExecutorInput` we no longer need to execute `EthClientExecutorInput` for empty blocks, consuming unnecessary cycles.
When we are provided with a non-empty set of executor inputs we simply assert:
1. TrustedStateRoot == ExecutorInputs.first().state_anchor()
2. TrustedHeight <= ExecutorInputs.first().parent_header().number

Overall this simplifies the proof system and also the prover service to gather input data. Now we can simply ignore `SignedHeader` blobs and only focus on `SignedData` blobs.

PR #96, opened 2025-07-17
refactor: block exec program equivocation and handling empty blocks

damiannolan:
## Overview

The following PR implements the equivocation tolerance as in discussed in #89, as well as the allowance for empty celestia blocks, ensuring that we can continue to aggregate block execution proofs in the event that a rollup application misses a celestia block #94. 

closes: #89 #94


damiannolan:
I wanted to generate real proofs using the prover network to add to the testdata and also run the aggregation program, however it seems that the prover network no longer support version 4.

```
2025-07-17T14:00:16.886945Z ERROR Permanent error encountered when requesting proof: unsupported prover version (Client specified an invalid argument)
```

crates/sp1/evm-exec/program/src/main.rs
```diff
@@ -39,73 +53,134 @@ use tendermint::block::Header;
...
+    if signed_data.len() != headers.len() {
+        let mut seen = HashSet::<u64>::new();
+        signed_data.retain(|sd| signed_data_height(sd).map(|h| seen.insert(h)).unwrap_or(false));
+    }
+
+    assert_eq!(
+        signed_data.len(),
+        headers.len(),
+        "Headers and SignedData must be of equal length"
+    );
```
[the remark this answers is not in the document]
damiannolan: I would propose doing this in a follow up PR as this is already large enough

crates/sp1/evm-exec/script/src/bin/datagen.rs
damiannolan, line 1: This script actually simplifies a good bit now, which is nice

crates/sp1/evm-exec/script/src/bin/main.rs
```diff
@@ -41,6 +43,12 @@ struct Args {
 
     #[arg(long)]
     height: u64,
+
+    #[arg(long)]
+    trusted_height: Option<u64>,
+
+    #[arg(long)]
+    trusted_root: Option<String>,
```
damiannolan, line 51: I still need to document why I added these, but that shouldn't stop PR review. 

Essentially, a celestia block may have zero EthClientExecutorInputs for evm blocks in it. In that event we can't just take the parent headers root/height in this script so best to allow via flags.

damiannolan:
Heads up, I think I will probably revise the CI workflow to not execute the programs and instead maybe add that functionality as workflow_dispatch, which can be done manually through the actions tab.

rootulp (approved):

crates/sp1/evm-exec/script/src/bin/main.rs
damiannolan: Done

PR #103, opened 2025-07-20
feat: add namespace and sequencer signature verification to block execution outputs

### LINKED ISSUE #95: Add sequencer signature verification to sp1 block execution program
## Overview 

As per discussion with @cmwaters, we will add sequencer signature verification to the block execution program.

This requires accepting the `PublicKey` as input to the program, verifying the signature in the `SignedData` and committing the `PublicKey` as a program output. 

It appears that the `signature` is over the `data` in:
```golang
message SignedData {
  Data data = 1;
  bytes signature = 2;
  Signer signer = 3;
}
```

ref: https://github.com/rollkit/rollkit/blob/main/block/manager.go#L994-L1004

### LINKED ISSUE #97: Add namespace to block execution program outputs
## Overview

We need to add the Namespace to the block execution program outputs.
This should be a pretty easy lift, could be done side by side with #95

tosynthegeek:
This PR addresses two related issues to the block execution program:

1. **Add namespace to block execution program outputs** (#97)
2. **Add sequencer signature verification** (#95)

### Namespace Addition (#97)
- Added `namespace: Namespace` field to `BlockExecOutput` struct
- Updated output creation in main execution program to include the namespace that was verified
- The namespace is already available as input and used for verification, now it's also committed as output

### Sequencer Signature Verification (#95)
- Added `public_key: Vec<u8>` field to `BlockExecOutput` struct  
- Added public key as input to the SP1 program via `sp1_zkvm::io::read_vec()`
- Implemented signature verification logic that:
  - Validates the signer's public key in `SignedData` matches the expected sequencer public key
  - Verifies the signature over the protobuf-encoded `Data` using Ed25519
- Added `ed25519-dalek = "2"` dependency for signature verification
- Both public key validation and signature verification occur in the blob equivalency section

Closes #97
Closes #95


damiannolan (approved):
Thanks! ❤️ 

I can regen some real proofs later on for the testdata directory.

tosynthegeek:
> Thanks! ❤️ I addressed comments and follow ups.
> 
> I can regen some real proofs later on for the testdata directory.

Sure ser. Would also look to pick up another issue😊

PR #108, opened 2025-07-22
chore: CLI tool for evm transactions

damiannolan:
## Overview

Adds a CLI tool for spamming basic EVM transactions so that we can gather various proof input data.


txflood/cmd/txflood/cmd/cmd.go
```diff
@@ -0,0 +1,165 @@
...
+	"time"
+
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/spf13/cobra"
+)
+
+var (
+	accountsFile = "accounts.json"
+	rpcURL       = "http://localhost:8545"
+)
```
damiannolan, line 18: Ideally would probably turn this into a config file or something but this is fine and bare minimum works for now

rootulp (approved):

Issue #110, opened 2025-07-24
Initial benchmarks for block execution program


damiannolan:
We want to record some initial benchmarks for the block execution progaram: `evm-exec`

Per Celestia Block Proof Benchmarks:
- [x] 0 blobs in namespace
- [x] 1 `SignedData` blob, 1 txs
- [x] 1 `SignedData` blob, 10 txs
- [x] 1 `SignedData` blob, 50 txs
- [x] 1 `SignedData` blob, 100 txs

We are primarily interested in gas and cycle count.

PR #111, opened 2025-07-24
test: add benchmark reports to evm-exec program

damiannolan:
## Overview

I spent some time playing with rust criterion, one of the standard benchmarking tools in rust. 
But after sometime, I figured it did not make sense for this type of program. For example, it requires a minimum sample size on each run of 10, and when we execute an SP1 program with a given set of inputs, the results are deterministic (gas, cycle count, etc).

This PR adds correct report cycle tracking for each step of the program and optionally can write this to a JSON output file. Using the JSON data we can create human readable charts using a simple python script.

Adds per Celestia Block Proof Benchmarks:
- [x] 0 blobs in namespace
- [x] 1 `SignedData` blob, 1 txs
- [x] 1 `SignedData` blob, 10 txs
- [x] 1 `SignedData` blob, 50 txs
- [x] 1 `SignedData` blob, 100 txs

See https://hackmd.io/@damian666/Sk59AvC8lg for markdown doc including the screenshots and some minor additional info on benchmarks added in this PR.

closes: #110 
closes: #112


rootulp (approved):
Super cool! Thanks for sharing the hackmd and including the Python tool to generate the charts

Cargo.toml
```diff
@@ -46,4 +50,12 @@ sp1-sdk = "=5.0.8"
 sp1-zkvm = "=5.0.8"
 tendermint = "0.40.1"
 tendermint-proto = "*"
-ed25519-dalek = "2"
+
+[patch.crates-io]
+# Add support for precompiles: https://docs.succinct.xyz/docs/sp1/optimizing-programs/precompiles
```
damiannolan, line 55: The difference when the precompiles are added is staggering

damiannolan:
Going to merge this so I can move forward with other benchmarks today. 

I've also added number of blobs in namespace and total evm block executions to the chart summary

Issue #86, opened 2025-07-03
Subscription based proof generation


damiannolan:
Proofs generated by the `evm-exec` program can be generated on a per-block, subscription basis and stored for later use.

This design allows a relayer to later request a range or aggregation proof on demand and leverage the previously generated per-block proofs to construct a path from a trusted height to a future height n, without needing to regenerate intermediate proofs.

PR #129, opened 2025-08-14
refactor: subscription based proof generation for evm-prover

damiannolan:
## Overview

WIP, will update desc

closes: #86

### DIFF
--- a/crates/evm-prover/src/prover/prover.rs
+++ b/crates/evm-prover/src/prover/prover.rs
@@ -5,19 +5,16 @@ use std::sync::Arc;
 
 use alloy_genesis::Genesis as AlloyGenesis;
 use alloy_provider::ProviderBuilder;
-use anyhow::{anyhow, bail, Context, Result};
+use anyhow::{anyhow, Context, Result};
 use async_trait::async_trait;
+use bytes::Bytes;
 use celestia_rpc::{client::Client, BlobClient, HeaderClient, ShareClient};
-use celestia_types::{nmt::Namespace, Blob, ExtendedHeader, ShareProof};
-use eq_common::KeccakInclusionToDataRootProofInput;
-use ev_types::v1::SignedHeader;
-use ev_types::v1::{store_service_client::StoreServiceClient, GetMetadataRequest, SignedData};
-use nmt_rs::{
-    simple_merkle::{db::MemDb, proof::Proof, tree::MerkleTree},
-    TmSha2Hasher,
-};
+use celestia_types::nmt::{Namespace, NamespaceProof};
+use celestia_types::{Blob, DataAvailabilityHeader, ExtendedHeader};
+use ev_types::v1::SignedData;
 use prost::Message;
 use reth_chainspec::ChainSpec;
+use rsp_client_executor::io::EthClientExecutorInput;
 use rsp_host_executor::EthHostExecutor;
 use rsp_primitives::genesis::Genesis;
 use rsp_rpc_db::RpcDb;
@@ -27,10 +24,6 @@ use sp1_sdk::{
     SP1VerifyingKey,
 };
 use tendermint::block::Header;
-use tendermint_proto::{
-    v0_38::{types::BlockId as RawBlockId, version::Consensus as RawConsensusVersion},
-    Protobuf,
-};
 
 use crate::config::config::{Config, APP_HOME, CONFIG_DIR, GENESIS_FILE};
 
@@ -170,6 +163,47 @@ impl BlockExecProver {
         }
     }
 
+    pub async fn start(&self) -> Result<()> {
+        let client = Client::new(&self.app.celestia_rpc, None).await?;
+        let mut subscription = client.blob_subscribe(self.app.namespace).await?;
+
+        while let Some(result) = subscription.next().await {
+            match result {
+                Ok(event) => {
+                    let blobs = event.blobs.unwrap_or_default();
+
+                    println!(
+                        "Received blob event for height: {}, blobs: {}",
+                        event.height,
+                        blobs.len()
+                    );
+
+                    let mut extended_header = client.header_get_by_height(event.height).await?;
+                    extended_header.header.version.app = 3; // TODO: need rs support for newer celestia app versions
+
+                    let namespace_data = client
+                        .share_get_namespace_data(&extended_header, self.app.namespace)
+                        .await?;
+
+                    let _proofs: Vec<NamespaceProof> =
+                        namespace_data.rows.iter().map(|row| row.proof.clone()).collect();
+
+                    // Determine EthClientExecutorInputs for celestia block
+                    let _signed_data: Vec<SignedData> = blobs
+                        .into_iter()
+                        .filter_map(|blob| SignedData::decode(Bytes::from(blob.data)).ok())
+                        .collect();
+                }
+                Err(e) => {
+                    eprintln!("Subscription error: {e}");
+                    break;
+                }
+            }
+        }
+
+        Ok(())
+    }
+
     /// Generates the serialized state transition function (STF) input for a given EVM block number.
     async fn generate_stf(&self, block_number: u64) -> Result<Vec<u8>> {
         let host_executor = EthHostExecutor::eth(self.app.chain_spec.clone(), None);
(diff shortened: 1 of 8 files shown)


crates/evm-prover/src/prover/prover.rs
```diff
@@ -141,6 +103,22 @@ impl AppContext {
...
+
+impl ProofJob {
+    fn new(height: u64, blobs: Vec<Blob>) -> Self {
+        Self { height, blobs }
+    }
+}
+
+// TODO: Add these as fields to the BlockExecProver to make configurable?
+const QUEUE_CAP: usize = 256;
+const CONCURRENCY: usize = 16;
```
damiannolan, line 120: We should add these settings to config to allow them to be tweaked if needs be. We can create an issue for this and do in a follow up.

crates/evm-prover/src/prover/prover.rs
```diff
@@ -100,29 +50,41 @@ pub struct AppContext {
...
                 .try_into()
                 .map_err(|e| anyhow!("Failed to convert genesis to chain spec: {e}"))?,
         );
 
         let raw_ns = hex::decode(config.namespace_hex)?;
         let namespace = Namespace::new_v0(raw_ns.as_ref()).context("Failed to construct Namespace")?;
+        let pub_key = hex::decode(config.pub_key)?;
+        let trusted_state = RwLock::new(TrustedState {
+            height: 0,
+            root: chain_spec.genesis_hash(),
```
damiannolan, line 77: The prover service currently starts with the evm app's genesis hash. 
Alternatively, we can query the latest from the node on startup and use that + its height instead. 

Any thoughts?

crates/evm-prover/src/prover/prover.rs
```diff
@@ -170,231 +186,164 @@ impl BlockExecProver {
...
+            namespace: self.app.namespace,
+            proofs,
+            executor_inputs,
+            trusted_height,
+            trusted_root,
+        };
 
-        Ok(stdin)
-    }
+        // TODO: store proofs for later use by aggregation circuit.
```
damiannolan, line 339: Will create an issue :)

crates/evm-prover/src/prover/prover.rs
```diff
@@ -413,93 +362,47 @@ pub struct BlockRangeExecProver {
...
+/// ProofInput is a convienience type used for proof aggregation inputs within the BlockRangeExecProver program.
+pub struct ProofInput {
+    proof: SP1Proof,
+    vkey: SP1VerifyingKey,
 }
 
 #[async_trait]
 impl ProgramProver for BlockRangeExecProver {
-    type Input = BlockRangeExecInput;
+    type Input = (BlockRangeExecInput, Vec<ProofInput>);
```
damiannolan, line 373: The input used here is now a tuple of the inputs we write via `stdin.write()` and the proofs we need to write with `stdin.write_proof()`. 

Putting the SP1 proof and vkey types onto the input struct we have in the lib crate `evm-exec-types` was causing me issues before with the compiler (I think that's why I avoid making those input types previously, cc. @jonas089).

I think using this tuple type works well here tho as there is a little bit of nuance involved between the regular input stream and the proof input stream.

testnet/ev-node/entrypoint.sh
```diff
@@ -5,44 +5,45 @@ cd /usr/bin
 
 sleep 5
 
-# Create default rollkit config if missing
+# Create default evnode config if missing
+# TODO: The --evnode.signer.path flag is not respected: https://github.com/evstack/ev-node/issues/2603
```
damiannolan, line 9: Once there is a fix available for this we can have a reproducible pubkey for the sequencer in our docker network

crates/evm-prover/src/grpc/server.rs
```diff
@@ -17,6 +21,21 @@ pub async fn create_grpc_server(config: Config) -> Result<()> {
         .build()
         .unwrap();
 
+    // TODO: Remove this config cloning when we can rely on the public key from config
+    // https://github.com/evstack/ev-node/issues/2603
+    let mut config_clone = config.clone();
+    config_clone.pub_key = public_key().await?;
+    println!("Successfully got pubkey from evnode: {}", config_clone.pub_key);
+
+    tokio::spawn({
```
damiannolan, line 30: I think we should rename this crate from grpc -> server and add shutdown cancellation/signal handling properly to these long lived tasks. Can do this in a follow up

damiannolan (commented):
Leaving some context for review before publishing the PR!

crates/evm-prover/src/prover/prover.rs
Manav-Aggarwal: I think ideal would be to query the reth node on startup for height 0, but seems like a low priority thing to do

damiannolan: Agreed, it is low priority for now. Querying for height 0 would yield the same state root available in the genesis. 
And I just realised as I wrote this response that I actually put the wrong value into this field, as genesis hash != genesis state root.

What I want here is `chain_spec.genesis_header().state_root`, I will update! 🙏🏻

crates/evm-prover/src/grpc/server.rs
damiannolan: https://github.com/celestiaorg/celestia-zkevm-hl-testnet/issues/152

testnet/ev-node/entrypoint.sh
damiannolan: #153

crates/evm-prover/src/prover/prover.rs
```diff
@@ -170,231 +191,164 @@ impl BlockExecProver {
...
 
-#[async_trait]
-impl ProgramProver for BlockExecProver {
-    type Input = BlockExecInput;
-    type Output = BlockExecOutput;
+        // NOTE: The following is an optimistic update to the tracked state. If we fail to produce a proof for
+        // the given inputs then we should shutdown the service, as developer intervention is likely required.
+        if let Some(next) = executor_inputs.last() {
+            let new_height = next.current_block.number;
+            let new_state_root = next.current_block.state_root;
```
blasrodri, line 325:
```suggestion
            let new_state_root = next.current_block.state_root;
            // Only update if moving forward
            // Race Scenario:

            // Time 1: Worker A starts (will end at EVM block 102)
            // Time 2: Worker B starts (will end at EVM block 105)
            // Time 3: Worker B finishes first → trusted_state = {height: 105, root: 0xBBB}
            // Time 4: Worker A finishes second → trusted_state = {height: 102, root: 0xAAA}

            // The Issue:

            // We went backwards from block 105 to block 102. The system now thinks it's at block 102 when it should be at block 105.

            // Why It's Bad:

            // Future proofs will use block 102's state as "trusted" when the real latest state is block 105. This breaks the proof chain.
            // It's important that this is checked inside the lock
            let mut s = self.app.trusted_state.write().await;
            if new_height > trusted_height {
                s.height = new_height;
                s.root = new_state_root;
            }
```

jonas089 (approved):

crates/evm-prover/src/prover/prover.rs
jonas089: I think we can do both, use the input struct and write the SP1 type values separately. Alternatively we can maybe serialize and deserialize them manually?

jonas089: Last time I did something like this I just serialized to raw bytes and then deserialized the struct in the circuit.

crates/evm-prover/src/prover/prover.rs
blasrodri: @damiannolan this seems to be a bug. So I suggest considering it, and including it in this PR.

damiannolan: You're 100% correct, there is definitely a race here. I've been going over this trying to iron out concretely what we want here.

These are the properties we are trying to achieve:
- Proof generation runs in parallel.
- Every event (height) produces a proof (for archival(store in db)/aggregation), even if later heights finish first.
- Each proof uses the trusted snapshot that was valid at the moment we scheduled that height.
- The global trusted_state (height, root) only advances sequentially in height order - never regresses - and does so before we start proving, so later jobs can immediately build on it.

Given these properties, I think we should implement a 3 stage pipeline, essentially lifting this locking out of its current location as well as the gathering of inputs.
1. Prep inputs (can be done in parallel). Gather proof inputs from rpc.
2. Assign (single-threaded, in-order, writes trusted_state). Gathers the trusted_height and trusted_root from state.
3. Prove (parallel). Take the assigned job and run prove() and store the result.

I have something working rn locally but will work on cleaning up the code before I push it here.

damiannolan: Refactored in https://github.com/celestiaorg/celestia-zkevm-hl-testnet/pull/129/commits/c1247d49dc7d5b208e886e6fc9689e1ff7407d29. I think this approach avoids races. I have a couple of TODOs left to complete. I can get to them on Monday and then get this PR merged

PR #162, opened 2025-09-01
Feat: EVM Merkle Proofs for 32 Byte Slots

jonas089:
This PR enables us to obtain and verify storage proofs against an EVM state root.
The incremental tree circuit will depend on this to verify the latest branch of the tree stored on chain (all 32 nodes of that branch).

We obtain the following proof

single account proof -> [storage_proof;TREE_DEPTH]

We verify the account proof against the state_root from the zk light client (for a block height `H`) and each of the TREE_DEPTH (likely = 32) against the account root.

We use alloy for RLP encoding/decoding. 

How to test:

Start the docker container:
```
make start
```

Submit as many Transactions as you like (at least 1):
```
make transfer
make transfer-back
```

Run the test to get and verify merkle proofs for all branches:

```
cargo test test_all_nodes_hyperlane_tree_branch --features alloy
```

But before that set the correct (latest) height here:

```
    #[tokio::test]
    async fn test_all_nodes_hyperlane_tree_branch() {
...
        let height = 1337;

        let proof = client
            .get_proof(
                // starts at 151 up to 182, count is located at 183
                // get the first one to check against off-chain tree
                &HYPERLANE_MERKLE_TREE_KEYS,
                contract,
                height,
            )
            .await
            .unwrap();
...
    }
```


crates/evm-storage-proofs/src/types.rs
```diff
@@ -0,0 +1,133 @@
+use alloy_primitives::{
+    Address, Bytes, FixedBytes, Uint,
+    hex::{FromHex, ToHexExt},
+};
+use alloy_rpc_types::EIP1186AccountProofResponse;
+use alloy_trie::{Nibbles, TrieAccount, proof::verify_proof};
```
damiannolan, line 6: whats a Nibbles?

crates/evm-storage-proofs/src/types.rs
```diff
@@ -0,0 +1,133 @@
...
+
+use crate::digest_keccak;
+
+pub const HYPERLANE_MERKLE_TREE_KEYS: [&str; 32] = [
```
damiannolan, line 10: where do these values come from? are they supposed to be the same as https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/solidity/contracts/libs/Merkle.sol#L173

crates/evm-storage-proofs/src/types.rs
```diff
@@ -0,0 +1,133 @@
...
+
+pub struct HyperlaneBranchProof {
+    pub proof: EIP1186AccountProofResponse,
+}
+
+impl HyperlaneBranchProof {
+    pub fn new(proof: EIP1186AccountProofResponse) -> Self {
+        Self { proof }
+    }
+    pub fn get_branch_node(&self, index: usize) -> String {
```
damiannolan, line 53: want to add some line breaks between functions for readability?

crates/evm-storage-proofs/src/types.rs
```diff
@@ -0,0 +1,133 @@
...
+    "0x00000000000000000000000000000000000000000000000000000000000000b0",
+    "0x00000000000000000000000000000000000000000000000000000000000000b1",
+    "0x00000000000000000000000000000000000000000000000000000000000000b2",
+    "0x00000000000000000000000000000000000000000000000000000000000000b3",
+    "0x00000000000000000000000000000000000000000000000000000000000000b4",
+    "0x00000000000000000000000000000000000000000000000000000000000000b5",
+    "0x00000000000000000000000000000000000000000000000000000000000000b6",
+];
+
+pub struct HyperlaneBranchProof {
```
damiannolan, line 45: It would be nice to have some doc comments on the struct and impl functions! :)

crates/evm-storage-proofs/src/client.rs
```diff
@@ -0,0 +1,119 @@
...
+            .provider
+            .get_proof(
+                contract,
+                keys.iter().map(|k| FixedBytes::from_hex(k).unwrap()).collect(),
+            )
+            .block_id(height.into())
+            .await?;
+        Ok(proof)
+    }
+    pub async fn get_storage_root(&self, height: u64) -> Result<String> {
```
damiannolan, line 46: should we call this `get_state_root`?

crates/evm-storage-proofs/src/client.rs
```diff
@@ -0,0 +1,119 @@
...
+
+pub struct EvmClient {
+    pub provider: DefaultProvider,
+}
+
+impl EvmClient {
+    pub fn new(provider: DefaultProvider) -> Self {
+        Self { provider }
+    }
+    pub async fn get_proof(
```
damiannolan, line 30: would also suggest line breaks and doc comments for this type too

crates/evm-storage-proofs/src/client.rs
```diff
@@ -0,0 +1,119 @@
+use alloy::{providers::Provider, rpc::types::EIP1186AccountProofResponse};
+use alloy_primitives::{Address, FixedBytes, hex::FromHex};
+use anyhow::{Context, Result};
+
+pub type DefaultProvider = alloy::providers::fillers::FillProvider<
```
damiannolan, line 5: I remember stepping through this chain of types before, its kinda mind blowing 🤯 

good call on a type alias lol

damiannolan (commented):
I just left some initial suggestions on code formatting and doc comments and a question on the tree keys. 

Otherwise lgtm!

crates/evm-storage-proofs/src/types.rs
jonas089: I'm surprised cargo fmt didn't 😮

crates/evm-storage-proofs/src/types.rs
jonas089: Will add a comment, I totally see where the confusion came from. Dw you're not stupid at all :)

crates/evm-storage-proofs/src/types.rs
jonas089: Haha it's one of those things you only see in Ethereum land. 
They use nibbles to encode keys in their Patricia Trie:
https://en.wikipedia.org/wiki/Nibble

Luckily Alloy takes care of these encodings for us, otherwise would be a big pain.

crates/evm-storage-proofs/src/types.rs
jonas089: Done 👍🏻

crates/evm-storage-proofs/src/client.rs
jonas089: done 👍🏻

crates/evm-storage-proofs/src/types.rs
jonas089: done 👍🏻

crates/evm-storage-proofs/src/types.rs
jonas089: done 👍🏻

PR #159, opened 2025-08-29
Feat: Hyperlane Event Indexer and Rocks DB

jonas089:
This PR implements an event indexer that can query all Hyperlane **Dispatch** messages for a given block range (in the test I use 0 to 10000). The **Dispatch** message is emitted from Hyperlane's **Mailbox** contract.

Additionally it implements all utilities to decode Hyperlane Message Headers and Bodies in Rust.

# New Crates Overview

`storage`: RocksDB implementation + simple interface
`evm-state-types`: Types required for Events and Hyperlane Messages, will be extended to also include the types required for the storage proof of the latest branch on-chain (this will be added in a future PR).
`evm-state-queries`: The event indexer itself, that takes an Arc<RwLock<HyperlaneMessageStore>> as input, as well as a websocket url and contract address (default is local testnet, otherwise use .env).

# Environment
New environment variables:

- HYPERLANE_MESSAGE_STORE="testdata/messages/hyperlane"
- RETH_WS_URL
- MAILBOX_CONTRACT_ADDRESS

# Full test setup
## 1. Run docker
`make start`

## 2. Submit transactions
```
make transfer
make transfer-back
```

## 3. run the indexer test
```
cargo test test_run_indexer -- --nocapture
```

The indexer will insert each new **Dispatch** message into the local RocksDB database. A test to decode and print the first message in the local db can be found in `evm-state-queries/src/hyperlane/mod.rs`. 

## Benchmarks and TPS
https://github.com/celestiaorg/hyperlane-tree-bench

PR #165, opened 2025-09-01
feat: add RocksDB storage for block-exec prover proofs

### LINKED ISSUE #154: Add storage for proofs created by the block-exec prover
Store the proofs generated by the block-exec prover in a db for later use.

I propose we use an embedded db such as rocksdb, but we are open to use whatever works best for the job. 
It will be write heavy, as a proofs will constantly be generated for each new block produced.

_Originally posted by @damiannolan in https://github.com/celestiaorg/celestia-zkevm-hl-testnet/pull/129#discussion_r2304060443_

blasrodri:
- Add ProofStorage trait with RocksDB implementation
- Store block proofs, range proofs, and Hyperlane messages
- Integrate automatic proof storage with BlockExecProver
- Add configurable storage path via Config.storage_path
- Include comprehensive unit tests and error handling
- Support efficient queries by height ranges and latest proof

closes #154


crates/evm-prover/src/storage/storage.rs
```diff
@@ -0,0 +1,301 @@
...
+    pub new_height: u64,
+    pub new_state_root: [u8; 32],
+    pub proof_data: Vec<u8>,
+    pub public_values: Vec<u8>,
+    pub created_at: u64,
+}
+
+
+#[async_trait]
+pub trait ProofStorage: Send + Sync {
```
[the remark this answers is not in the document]
blasrodri: I was thinking more in terms of potentially mocking this, or being flexible if we wanted sqlite, etc. But happy to remove the trait if not needed.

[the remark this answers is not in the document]
jonas089: No need to remove it, is this storage module going to be separate from the one I created in the other PR?

https://github.com/celestiaorg/celestia-zkevm-hl-testnet/pull/159

Is it a good idea to merge them into a single module or not so much?

blasrodri:
> I think you'll need to change the base branch to `main`

I thought it had to be based on your branch

crates/evm-prover/src/prover/prover.rs
```diff
@@ -191,11 +193,38 @@ struct ScheduledProofJob {
...
         let config = BlockExecProver::default_config();
         let prover = ProverClient::from_env();
 
-        Arc::new(Self { app, config, prover })
+        // Initialize RocksDB storage in the default data directory
+        let storage_path = dirs::home_dir()
+            .expect("cannot find home directory")
+            .join(APP_HOME)
+            .join("data")
+            .join("proofs.db");
```
damiannolan, line 205: 🙏🏻

crates/evm-prover/src/prover/prover.rs
```diff
@@ -447,11 +476,21 @@ impl BlockExecProver {
             trusted_root: scheduled.trusted_root,
         };
 
-        // TODO: Add storage for SP1ProofWithPublicValues: https://github.com/celestiaorg/celestia-zkevm-hl-testnet/issues/154
-        let (_proof, outputs) = self.prove(inputs).await?;
+        let (proof, outputs) = self.prove(inputs).await?;
+
+        // Store the proof in the database
+        if let Err(_e) = self
```
damiannolan, line 482: Maybe we shouldn't discard `e` and log it below in the vent of an error?

crates/evm-prover/src/prover/prover.rs
```diff
@@ -447,11 +476,21 @@ impl BlockExecProver {
             trusted_root: scheduled.trusted_root,
         };
 
-        // TODO: Add storage for SP1ProofWithPublicValues: https://github.com/celestiaorg/celestia-zkevm-hl-testnet/issues/154
-        let (_proof, outputs) = self.prove(inputs).await?;
+        let (proof, outputs) = self.prove(inputs).await?;
+
+        // Store the proof in the database
```
damiannolan, line 481: I think we can remove this comment, I don't think it provides much value

crates/evm-prover/src/storage/storage.rs
```diff
@@ -0,0 +1,300 @@
...
+pub struct StoredBlockProof {
+    pub celestia_height: u64,
+    pub celestia_header_hash: [u8; 32],
+    pub evm_height: u64,
+    pub evm_state_root: [u8; 32],
+    pub namespace: Namespace,
+    pub proof_data: Vec<u8>,
+    pub public_values: Vec<u8>,
+    pub created_at: u64,
+}
```
damiannolan, line 36: I think we can reduce this as many of the fields are duplicates since they are encoded within `public_values`:

```suggestion
pub struct StoredBlockProof {
    pub celestia_height: u64,
    pub proof_data: Vec<u8>,
    pub public_values: Vec<u8>,
    pub created_at: u64,
}
```

crates/evm-prover/src/storage/storage.rs
```diff
@@ -0,0 +1,300 @@
...
+    pub end_height: u64,
+    pub celestia_header_hash: [u8; 32],
+    pub trusted_height: u64,
+    pub trusted_state_root: [u8; 32],
+    pub new_height: u64,
+    pub new_state_root: [u8; 32],
+    pub proof_data: Vec<u8>,
+    pub public_values: Vec<u8>,
+    pub created_at: u64,
+}
```
damiannolan, line 51: Similarly, I think we can probably do the same for this struct and just keep a subset of the fields (not sure what `id` is or if its needed):

```suggestion
pub struct StoredRangeProof {
    pub start_height: u64,
    pub end_height: u64,
    pub proof_data: Vec<u8>,
    pub public_values: Vec<u8>,
    pub created_at: u64,
}
```

I assume we create the `range_key` as a composite with start/end height! 👍🏻

crates/evm-prover/examples/proof_storage_demo.rs
damiannolan, line 1: I think we can probably omit this file as the unit tests have things covered!

damiannolan (commented):
Excellent work so far, I just dropped some comments on structs..etc

After this PR when we have our proofs stored we can start thinking about wiring up the impl for the grpc service: https://github.com/celestiaorg/celestia-zkevm-hl-testnet/blob/main/crates/evm-prover/src/prover/service.rs#L17-L23

The goal there would be the to query the store for a particular range requested by the client, and generate a proof using the `BlockRangeExecProver`. We can create an issue for that!

crates/evm-prover/src/storage/storage.rs
```diff
@@ -0,0 +1,265 @@
...
+        end_height: u64,
+    ) -> Result<Vec<StoredBlockProof>, ProofStorageError>;
+
+    #[allow(dead_code)]
+    async fn get_latest_block_proof(&self) -> Result<Option<StoredBlockProof>, ProofStorageError>;
+}
+
+pub struct RocksDbProofStorage {
+    db: Arc<DB>,
+}
```
damiannolan, line 84: When Jonas has finished his work creating the circuit state membership/inclusion we will need to integrate it into the prover service also. 

Should we be sharing a single `Arc<DB>` for all storage and have data arranged by column families (as you have already done here)?

I assume the approach would be to just add a new `CF_MESSAGES` for hyperlane messages, rather than creating another struct with a db field.

damiannolan (approved):
Thanks for addressing the comments. Awesome work @blasrodri, nice one! 🙏🏻

PR #169, opened 2025-09-02
refactor: avoid base64 encoding when serialising namespace outputs

damiannolan:
## Overview

WIP: don't review


jonas089 (approved):
lgtm

PR #170, opened 2025-09-02
chore: assert proofs len against number of row roots

damiannolan:
## Overview

Adding defensive check from code walkthrough


blasrodri (approved):

jonas089:
> Will need to regen vkeys before publishing as ready for review

Will approve once that's done :).

PR #172, opened 2025-09-03
dep: bump SP1 v5.2.1

jonas089:
This PR bumps our dependency on SP1 to the latest.

It also includes the check that we discussed for proof length vs. number of roots and all vks + proofs were re-generated.

@blasrodri @damiannolan please run:

`sp1up --version 5.2.1` 

so that rust-analyzer doesn't get confused and everything compiles.


crates/sp1/evm-exec/program/src/main.rs
```diff
@@ -84,12 +84,14 @@ pub fn main() {
         }
     }
 
+    assert_eq!(
```
damiannolan, line 87: 🙏🏻

damiannolan (approved):
Good stuff! Thanks. Do we need to update any of the sp1 precompile patches as well?

jonas089:
> Good stuff! Thanks. Do we need to update any of the sp1 precompile patches as well?

Nope those can stay as they are for now :).

PR #174, opened 2025-09-04
feat: Circuit to Prove Hyperlane Messages

jonas089:
This PR introduces a new Circuit capable of proving Hyperlane message inserts into the incremental Tree and verifying them against the on-chain state of said tree at a given EVM block height.

The core verification logic looks like this:
```rust
    pub fn verify(&mut self) {
        let message_ids: Vec<String> = self.messages.iter().map(|m| m.id()).collect();
        for message_id in message_ids {
            self.snapshot
                .insert(message_id)
                .expect("Failed to insert message id into snapshot");
        }
        assert!(self.branch_proof.verify(
            &HYPERLANE_MERKLE_TREE_KEYS,
            Address::from_str(&self.contract).unwrap(),
            &self.state_root,
        )?);
        for idx in 0..HYPERLANE_MERKLE_TREE_KEYS.len() {
            // The branch nodes of the snapshot after insert must match the branch nodes of the incremental
            // tree on the EVM chain.
            assert_eq!(self.snapshot.branch[idx], self.branch_proof.get_branch_node(idx));
        }
    }
```

Where a bunch of message ids  that were previously inserted into our local message store using the event indexer are inserted into an off-chain snapshot (the most recent one) of the Tree. Additionally, we obtain an account & storage proof for the `TREE_DEPTH` (=32) branch nodes that are stored on-chain.

The merkle proof must be valid against our current root for the MerkleTreeHook contract and the branch nodes must equal the branch nodes of our snapshot after inserting the new messages.

If both of these conditions are satisfied, then we know that the messages indeed occurred and were inserted by the time our `target_height` got finalized.

Run a test:

1. update the .env and specify a path for the message store:

```
HYPERLANE_MESSAGE_STORE="/Users/chef/Desktop/celestia-zkevm-hl-testnet/.db/messages/hyperlane"
```

2. run the docker container(s):
```
make start
```

Wait for the docker containers to spin up fully and proceed with 3.

3. Insert messages:

```
make transfer

make transfer-back
...
```

4. Run the indexer to store the messages

Currently we do this by running the test which will fetch all Hyperlane messages that occurred in the first 10000 blocks:

```
cargo test test_run_indexer -- --nocapture
```

Output:

```
Inserted Hyperlane Message at index: 61
Inserted Hyperlane Message at index: 62
Inserted Hyperlane Message at index: 63
Inserted Hyperlane Message at index: 64
Inserted Hyperlane Message at index: 65
Inserted Hyperlane Message at index: 66
[...]

PR #178, opened 2025-09-08
feat: Re-structure evm-prover and add implementation for Membership circuit

jonas089:
This PR is based on #174 and implements the Prover Trait for the Memership Circuit.

Additionally the AppContext is modified and wrapped in an Arc<RwLock> to be shared by multiple circuit's prover tasks.

@damiannolan @blasrodri


jonas089:
> > @blasrodri @damiannolan This is a draft but I'd like to ask for your opinion already.
> > I decided to wrap the AppContext in an Arc (Tokio RwLock) so that it can be shared between the two prover tasks for the two circuits (Block prover, Membership Prover).
> > Next I'd like to implement the prover trait for the Membership Circuit which will make it easier for us to run the E2E. I want to make sure though that we are all aligned on the architecture and that you agree it makes sense to wrap AppContext in an Arc for this purpose.
> 
> what do you need specifically to share within the `AppContext`? The only thing that can change is the `trusted_state`, right? Anything else can be copied and shared by both workers, as they won't be modifying it, is that correct?

I could just add the new shared state value that I'm introducing for my circuit to the trusted_state and wrap only that in an RwLock. 

The rest can just be shared references so yeah I'll revert that and only Arc<RwLock> the trusted_state. 
Wasn't sure what's better but since we only modify that I guess you're right.

PR #179, opened 2025-09-08
fix: Version lock ev-node and fix container init

jonas089:
This PR version-locks the ev-node in the Dockerfile to `ghcr.io/evstack/ev-node-evm-single@sha256:8e19df666dafc072a36924b46511f0016ef189e119f9f0d85c182e5c8dd67559`.

@damiannolan and I were on different versions causing mine to break due to an update in the ev-node API.
We can upgrade to future tags or releases later down the road, for now it is important that the environment is reproducible for us during development.

I ran the prover on all circuits and they still function as before, so no breaking changes expected.


damiannolan (approved):
Thank you @jonas089 🙏🏻

PR #196, opened 2025-09-18
Enable celestia API

leon-do:
## Overview

Enable API to connect to wallet like Leap or Cosmostation.

Screenshot 2025-09-17 at 8 41 45 PM


docker-compose.yml
```diff
@@ -26,7 +27,7 @@ services:
     volumes:
       - celestia-app:/home/celestia/.celestia-app
     command: >
-      start --force-no-bbr
+      start --force-no-bbr --api.enable true --api.address tcp://0.0.0.0:1317
```
damiannolan, line 30: Could we adapt this to rather than providing start flags, override the config directly in the `init.sh` used by the initialisation container?

See here: https://github.com/celestiaorg/celestia-zkevm-hl-testnet/blob/main/testnet/celestia-app/init.sh#L15-L17

Similarly to how we enable grpc, we can simply extend it like:

```bash
celestia-appd config set app api.enable true
celestia-appd config set app api.address tcp://0.0.0.0:1317
```

damiannolan (changes requested):
Good call on enabling the API server 👍🏻 

I just suggested a different way of doing it to align with how we are overriding config opts in the container setup, otherwise lgtm!

damiannolan:
> Updated. sorry about that.

No worries at all

damiannolan (approved):

PR #202, opened 2025-09-19
enhance: Improve crate structure by merging types into a single crate and clean up dependencies

jonas089:
This PR merges all types into a single crate `ev-zkevm-types` and maintains the logical structure / separation of concerns for the different programs and modules.

- The custom evm-client implementation was removed because it was just a wrapper around provider (we use provider everywhere now)
- Some type related crates got deleted (replaced by `ev-zkevm-types`)
- Several dependencies were moved to the workspace root and are now imported via `workspace = true`


jonas089:
#205 builds on top of this and refactors the storage implementation, this + 205 = clean workspace :)

crates/ev-zkevm-types/src/lib.rs
```diff
@@ -2,6 +2,7 @@ use serde::{Deserialize, Serialize};
 
 pub mod events;
 pub mod hyperlane;
+pub mod programs;
 use hyperlane::HyperlaneMessage;
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
```
damiannolan, line 8: Should these `StoredHyperlaneMessage` struct and impls be elsewhere? inside programs/hyperlane or something?

crates/ev-zkevm-types/src/hyperlane.rs
damiannolan, line 1: would image this could go into programs/hyperlane ?

crates/ev-zkevm-types/src/events.rs
damiannolan, line 1: This too? could move into programs/hyperlane ?

damiannolan (approved):
Looks good, thank you @jonas089! 

Largely in favour of the general cleanup, I think we could improve the structure of the new ev-zkevm-types crate a little bit tho with grouping of types. I left some comments. 
Not particularly blocking, we can clean this up later again, but if you would like to move things before merging then please do! 🙏🏻

crates/ev-zkevm-types/src/lib.rs
jonas089: I'll move it to the storage crate because that's the only place where it is used :).

Issue #211, opened 2025-09-23
Task: Submit Proof Payloads (Block Proof, Message Proof)


jonas089:
Take the proofs from #209  and #210, combine them into a single proof object/payload and submit them to the ISM for verification. Implement this as another method in our e2e crate so that it can be wrapped in a `make e2e` command o.e. easily.

Issue #209, opened 2025-09-23
Task: Prepare E2E payload for block proof


jonas089:
In our new `e2e` crate, implement a method to create a block proof on demand, such that it can be wrapped in a make command for `make e2e`, o.e.

The method should generate a block proof, recurse it using block-range-exec and return it.

PR #212, opened 2025-09-23
feat: E2E proof generation for the message circuit

jonas089:
Depends on #205 

This PR introduces the `e2e` crate that will contain all the logic for generating proofs and submitting them to the Celestia ISM.

This PR includes an implementation of the message prover that will generate part of the payload sent to the ISM.

A follow up PR will do the same for the block prover, the file where said function will live has been created but the implementation will be part of  #209.

Once this PR is merged and the future one that will address #209, we can open a PR that addresses #211.
At that point we will have successfully implemented the E2E (on the prover side). @damiannolan @Manav-Aggarwal @blasrodri


Manav-Aggarwal:
Great work :) Excited to see the E2E working

damiannolan (commented):
Overall from quick skim looks amazing! I will try to prioritise reviewing this soon! 🙏🏻

PR #215, opened 2025-09-24
critical: Fix block prover circuit

jonas089:
This PR partially implements the block prover for the E2E. 
A follow up PR will finish that implementation by wrapping the block proofs in a single groth16 proof using `ev-range-exec`.

Additionally and more importantly, this PR includes a critical fix:

The circuit logic for the block prover was flawed because we did not at all account for `absence` proofs. Hence the assertion "Number of proofs must equal the number of row roots" was triggered every time a root for an absence proof was included in the circuit inputs.

I temporarily fixed this by supplying the absence proofs as inputs and skipping them (this should be fine since all data that mutates state / all transactions that are being executed must be signed by the sequencer (we verify this) and the NameSpaceProofs only progress `count` when they are `presence`. Absence proofs don't progress `count`.

For now I think that we should merge this fix so that we can push for E2E, but we should definitely revisit this section to make sure that the absence proofs do not negatively impact the soundness of our block prover.

@damiannolan @blasrodri @Manav-Aggarwal


crates/e2e/README.md
```diff
@@ -0,0 +1,29 @@
...
+Take note of one of the block heights after transaction submission, so that you can update the `target_height` in the config file.
+
+Next update the configuration file in `e2e/config.rs`:
+
+In `e2e/config.rs`:
+```rust
+...
+pub const TARGET_HEIGHT: u64 = 250;
+```
+Make sure that before TARGET_HEIGHT transactions have occurred. The E2E will prove all transactions in range `0` to `TARGET_HEIGHT` in one go.
```
blasrodri, line 23: make sure that TARGET_HEIGHT have occurred before ...
seems like the wording got swapped

crates/ev-state-queries/src/hyperlane/indexer.rs
```diff
@@ -53,12 +53,11 @@ impl HyperlaneIndexer {
                 Ok(event) => {
                     let dispatch_event: DispatchEvent = event.into();
                     let current_index = message_store.current_index()?;
-                    let next_index = current_index;
                     let hyperlane_message =
                         decode_hyperlane_message(&dispatch_event.message).expect("Failed to decode Hyperlane message");
                     let stored_message = StoredHyperlaneMessage::new(hyperlane_message, log.block_number);
-                    message_store.insert_message(next_index, stored_message).unwrap();
-                    println!("Inserted Hyperlane Message at index: {next_index}");
+                    message_store.insert_message(current_index, stored_message).unwrap();
```
blasrodri, line 59: why this change?

docker-compose.yml
```diff
@@ -35,7 +35,7 @@ services:
       - celestia-zkevm-net
 
   celestia-bridge:
-    image: ghcr.io/celestiaorg/celestia-node:v0.27.0-arabica
+    image: ghcr.io/celestiaorg/celestia-node:v0.25.0-arabica
```
blasrodri, line 38: why downgrade?

crates/sp1/ev-exec/program/src/main.rs
```diff
@@ -106,8 +106,13 @@ pub fn main() {
 
     let mut cursor = 0;
     for (proof, root) in inputs.proofs.iter().zip(roots) {
-        assert!(proof.is_of_presence(), "NamespaceProof must be of presence");
-
+        // we also need to verify absence proofs because we get a root for them
+        if proof.is_of_absence() {
```
damiannolan, line 110: I introduced this issue in https://github.com/celestiaorg/celestia-zkevm-hl-testnet/pull/188

There is a chance that this case was being hit previously but we were calling `verify_complete_namespace()` with an empty set of leaves against the root which would have allowed it succeed, but unbeknownst to us realising it was an absence proof being verified. 

This actually seems to me like more or a problem with the contains()` method on the namespace ids: https://docs.rs/nmt-rs/latest/src/nmt_rs/namespaced_hash.rs.html#345-352

And a more fitting name would probably be `is_in_range()`

damiannolan (changes requested):
Can you lift the diffs for the absence proofs fix into a smaller scoped PR please? 🙏🏻

crates/ev-state-queries/src/hyperlane/indexer.rs
jonas089: I assigned next_index to be equal to current_index so next_index was a useless variable.

crates/e2e/src/bin/block.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[tokio::main]
+async fn main() {
+    dotenvy::dotenv().ok();
+    prove_blocks(
+        START_HEIGHT,
+        TRUSTED_HEIGHT,
+        NUM_BLOCKS,
+        &mut FixedBytes::from_hex(TRUSTED_ROOT).unwrap(),
+    )
```
damiannolan, line 15: As a follow up we might want to add an `Args` struct like we have for example in (ev-exec/script) and parse using clap;

Example pasted from ev-exec:
```rust
/// The arguments for the command.
#[derive(Parser, Debug)]
#[command(author, version = clap::crate_version!(), about = "A CLI for running the ev-exec SP1 program", long_about = None)]
struct Args {
    #[arg(long, help = "Run the program in execute mode")]
    execute: bool,

    #[arg(long, help = "Run the program in prove mode")]
    prove: bool,

    #[arg(long, help = "The Celestia block height")]
    height: u64,

    #[arg(long, help = "Output file for benchmark report in JSON format")]
    output_file: Option<String>,

    #[arg(long, help = "Trusted EVM height which contains trusted state root")]
    trusted_height: Option<u64>,

    #[arg(long, help = "Trusted state root (hex string) for the trusted height")]
    trusted_root: Option<String>,
}

```

crates/ev-prover/src/prover/programs/block.rs
```diff
@@ -56,8 +56,8 @@ pub struct AppContext {
 /// This type is wrapped in a RwLock by the AppContext such that it can be updated safely across concurrent tasks.
 /// Updates are made optimisticly using the EthClientExecutorInputs queried from the configured EVM full node.
 pub struct TrustedState {
-    height: u64,
-    root: FixedBytes<32>,
+    pub height: u64,
+    pub root: FixedBytes<32>,
```
damiannolan, line 60: Are these diffs required?

crates/ev-prover/src/commands/command.rs
```diff
@@ -39,6 +39,7 @@ pub fn init() -> Result<()> {
 }
 
 pub async fn start() -> Result<()> {
+    dotenvy::dotenv().ok();
```
damiannolan, line 42: is this required in ev-prover?

crates/e2e/src/prover/block.rs
```diff
@@ -1,10 +1,490 @@
...
+        if prover_mode == "network" {
+            parallel_prover(start_height, &mut trusted_height, num_blocks, trusted_root).await?
+        }
+        // mock mode is not possible for recursive groth16 proofs
+        else if prover_mode == "mock" {
+            panic!("Recursive groth16 proofs are not supported in mock mode");
+        }
+        // synchroneous mode (cuda, cpu)
+        else {
+            synchroneous_prover(start_height, &mut trusted_height, num_blocks, trusted_root).await?
```
damiannolan, line 104: misspelling here: synchroneous -> synchronous

crates/e2e/src/prover/block.rs
```diff
@@ -1,10 +1,490 @@
...
+
+        current_trusted_height = public_values.new_height;
+        current_trusted_root = public_values.new_state_root.into();
+        trusted_heights.push(public_values.new_height);
+        trusted_roots.push(public_values.new_state_root.into());
+    }
+
+    // now we can generate proofs in parallel
+    let mut handles: Vec<JoinHandle<()>> = Vec::new();
+    for block_number in start_height..(start_height + num_blocks) {
```
damiannolan, line 223: so because you run up front in `execute` mode your populate two vectors of trusted heights and roots to then pass into each as inputs when running in `prove` mode?

is that the gist here?

crates/ev-prover/src/commands/command.rs
jonas089: if we don't initialize dotenvy then I think it will always default to CPU prover e.g. not consider our SP1 env variables.

damiannolan: want to move it to the main() fn then?

jonas089: yes

crates/ev-prover/src/prover/programs/block.rs
jonas089: removed

crates/e2e/src/bin/block.rs
jonas089: So we implement a CLI for E2E to prove both circuits and pass the args to both prove functions? That works I think.

jonas089: Wouldn't the best case be to have the e2e work entirely without CLI, just my running make prove or something like that?

crates/e2e/src/prover/block.rs
```diff
@@ -1,10 +1,489 @@
...
+            .expect("Failed to get extended header");
+        let namespace_data = celestia_client
+            .share_get_namespace_data(&extended_header, namespace)
+            .await
+            .expect("Failed to get namespace data");
+        let mut proofs: Vec<NamespaceProof> = Vec::new();
+        for row in namespace_data.rows {
+            proofs.push(row.proof);
+        }
+        println!("Got NamespaceProofs, total: {}", proofs.len());
```
damiannolan, line 172: nit; could we add a little bit of "breathing space" for the code here, becomes easier to read when there's a few line breaks imo

```suggestion
        println!("\nProcessing block: {block_number}");
        
        let blobs: Vec<Blob> = celestia_client
            .blob_get_all(block_number, &[namespace])
            .await
            .expect("Failed to get blobs")
            .unwrap_or_default();
        
        println!("Got {} blobs for block: {}", blobs.len(), block_number);

        let extended_header = celestia_client
            .header_get_by_height(block_number)
            .await
            .expect("Failed to get extended header");
        
        let namespace_data = celestia_client
            .share_get_namespace_data(&extended_header, namespace)
            .await
            .expect("Failed to get namespace data");
        
        let mut proofs: Vec<NamespaceProof> = Vec::new();
        for row in namespace_data.rows {
            proofs.push(row.proof);
        }
        
        println!("Got NamespaceProofs, total: {}", proofs.len());
```

damiannolan (approved):
I think this is good to go right? Any follow up changes can be made later if required!

PR #219, opened 2025-09-29
chore: zk ism deployment override method

damiannolan:
## Overview

Deploys the zk ism on top of an existing noop stack - overwriting the ism on the mailbox and token

### DIFF
--- a/hyperlane/scripts/docker-entrypoint.sh
+++ b/hyperlane/scripts/docker-entrypoint.sh
@@ -16,8 +16,8 @@ if [[ ! -f "$CONFIG_FILE" ]]; then
   echo "Deploying Hyperlane warp synthetic token EVM contracts..."
   hyperlane warp deploy --config ./configs/warp-config.yaml --registry ./registry --yes
 
-  echo "Deploying Hyperlane on cosmosnative..."
-  hyp deploy-zkism celestia-validator:9090 reth:8545 ev-node-evm-single:7331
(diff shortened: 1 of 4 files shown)


hyperlane/scripts/docker-entrypoint.sh
```diff
@@ -34,6 +34,8 @@ if [[ ! -f "$CONFIG_FILE" ]]; then
 
   echo "Configuring remote router for warp route on cosmosnative..."
   hyp enroll-remote-router celestia-validator:9090 0x726f757465725f61707000000000000000000000000000010000000000000000 1234 0x000000000000000000000000345a583028762De4d733852c9D4f419077093A48
+
+  hyp setup-zkism celestia-validator:9090 reth:8545 ev-node-evm-single:7331
```
damiannolan, line 38: and then this will deploy the zk ism and overwrite the ids on the mailbox and token

.github/workflows/docker.yml
```diff
@@ -6,6 +6,7 @@ on:
     paths:
       - '**/Dockerfile'
       - '.github/workflows/docker.yml'
+      - 'hyperlane/**'
```
damiannolan, line 9: This will trigger the workflow to rebuild the image any time something in the hyperlane dir changes.

jonas089: makes sense

PR #222, opened 2025-10-02
Feat: Full round-trip E2E with Tia balance increase

jonas089:
This PR adds message submission to the Hyperlane core.

After submitting the block & message proof we want to send the raw message bytes as a `MsgSubmitMessages` to `hyperlane.core.v1` which will check that the message ID was included in the ZKISM and then unlock the assets on Celestia.

Since no ZKP is involved in this step it is as easy as sending the message to the module.

Next steps:

- get the trusted EVM height from the ZKISM and find the corresponding Celestia height (that Celestia height + 1 = START_HEIGHT) - DONE
- get the trusted EVM root from the ZKISM - PENDING
- run make transfer-back and query the transaction to find out which EVM block it was included in => that's our TARGET_HEIGHT (EVM) - PENDING
- find the corresponding Celestia block for TARGET_HEIGHT such that NUM_BLOCKS = that Celestia block - START_HEIGHT + 1 - DONE


crates/celestia-grpc-client/src/proto/celestia.zkism.v1.rs
```diff
@@ -0,0 +1,286 @@
...
+    pub signer: ::prost::alloc::string::String,
+}
+/// MsgSubmitMessagesResponse is the response type for SubmitMessages.
+#[allow(clippy::derive_partial_eq_without_eq)]
+#[derive(Clone, Copy, PartialEq, ::prost::Message)]
+pub struct MsgSubmitMessagesResponse {}
+/// MsgProcessMessage is the request type for ProcessMessage.
+#[allow(clippy::derive_partial_eq_without_eq)]
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct MsgProcessMessage {
```
damiannolan, line 253: This file is code-generated, we shouldn't be manually adding things here.

I have resolved conflicts and will push a change including this into the new build system

crates/celestia-grpc-client/src/proto/hyperlane.core.v1.rs
damiannolan, line 1: Added with:

```
buf generate --template buf.gen.yaml \
  buf.build/bcp-innovations/hyperlane-cosmos:v1.0.1 \
  --path hyperlane/core/v1/tx.proto
```

damiannolan: We only need the `tx.proto` types so I just limited it to this. We can include others in future if needs be

damiannolan: cc @blasrodri you could include the warp tx.proto similarly for enable the transfers via the same client

crates/e2e/src/bin/e2e.rs
```diff
@@ -89,55 +75,40 @@ async fn main() {
...
+        assert!(response.success);
     }
-
-    Err(anyhow::anyhow!("Timeout waiting for tx {tx_hash:?}"))
 }
 
-/*
-I use this command
+// todo: find a place for this function and remove it from the binaries
+async fn inclusion_height(block_number: u64) -> anyhow::Result<u64> {
```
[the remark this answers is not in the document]
jonas089: Yes, because I didn't want to put that code in any shared library between the two test binaries. Later I want to remove both and import from wherever we end up putting this function.

crates/e2e/src/bin/e2e.rs
```diff
@@ -28,53 +23,44 @@ async fn main() {
...
     let client: Arc<EnvProver> = Arc::new(ProverClient::from_env());
-    let tx_config = celestia_grpc::TxConfig {
-        gas_limit: Some(200000u64),
-        gas_price: Some(1000_f64),
-        memo: Some("zkISM state transition proof submission".to_string()),
-        ..Default::default()
-    };
+    let trusted_inclusion_height = inclusion_height(ism_trusted_height).await.unwrap() + 1;
+    let target_inclusion_height = inclusion_height(TARGET_HEIGHT).await.unwrap();
+    let num_blocks = target_inclusion_height - trusted_inclusion_height + 1;
```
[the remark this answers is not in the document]
jonas089: Because we prove in range 0..N where N is excluded => if we want to prove up to and including our target N we must add 1 to num_blocks.

Example:

Trusted Height: 5, Target Height: 10

Blocks proven: 5,6,7,8,9,10,11,12,13,14 (excluding our target block 15)

We don't have to prove the trusted block 5 though, so what I'll do is change it to exclude the first index and include the last index directly in the loop and then remove the +1.

crates/e2e/src/bin/e2e.rs
```diff
@@ -89,55 +75,40 @@ async fn main() {
...
-        if attempts > 12 {
-            return Err(anyhow::anyhow!("Timeout waiting for tx {tx_hash:?}"));
-        }
-        sleep(Duration::from_secs(5)).await;
+    // submit all now verified messages to hyperlane
+    for message in message_proof.1 {
+        let message_hex = alloy::hex::encode(encode_hyperlane_message(&message.message).unwrap());
+        let msg = MsgProcessMessage::new(
+            "0x68797065726c616e650000000000000000000000000000000000000000000000".to_string(),
+            "celestia1y3kf30y9zprqzr2g2gjjkw3wls0a35pfs3a58q".to_string(),
```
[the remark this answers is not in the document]
jonas089: done

PR #227, opened 2025-10-07
Feat: Tracing in Prover Service

jonas089:
This PR removes all print statements from the message and block prover service and replaces them with `tracing`.

The tracing subscriber is initialized in such a way that it filters all `Info` and `Debug` level logs for SP1 out, but shows `warning` and above. The debug level for the ev-prover can be specified using the environment variable, e.g. RUST_LOG="ev-prover=debug".

The tracing subscriber is subscribed only once at the instantiation of the grpc server that spawns the prover tasks.


jonas089:
I based this PR off #222, will rebase once #222 is merged.

damiannolan (approved):

damiannolan:
Did you intentionally close this PR? @jonas089

PR #230, opened 2025-10-07
chore: add hyperlane transfers to rust tx client

damiannolan:


crates/celestia-grpc-client/src/bin/client.rs
```diff
@@ -45,6 +47,22 @@ enum Commands {
         #[arg(long)]
         height: u64,
     },
+    Transfer {
```
damiannolan, line 50: update these doc strings

crates/celestia-grpc-client/src/client.rs
```diff
@@ -191,7 +192,7 @@ impl ProofSubmitter for CelestiaIsmClient {
             return Err(IsmClientError::InvalidProof("ISM ID cannot be empty".to_string()));
         }
 
-        self.send_tx(proof_msg, "MsgUpdateZKExecutionISM").await
+        self.send_tx(proof_msg).await
     }
 
     async fn submit_state_inclusion_proof(&self, proof_msg: StateInclusionProofMsg) -> Result<TxResponse> {
```
damiannolan, line 198: I don't think we need this `ProofSubmitter` trait, do we? I think we should consider removing it.
We can just horse the messages straight at the rpc with the `send_tx` method.

crates/celestia-grpc-client/src/client.rs
```diff
@@ -132,19 +132,20 @@ impl CelestiaIsmClient {
     }
 
     /// Sign and send a tx to Celestia including the provided message.
-    async fn send_tx<M>(&self, message: M, message_type: &str) -> Result<TxResponse>
```
damiannolan, line 135: I removed the `message_type` arg as it can be inferred from the type url

crates/celestia-grpc-client/src/client.rs
blasrodri: Yes, you're right. Not adding any value

blasrodri (approved):

PR #237, opened 2025-10-09
feat: add dedicated gRPC methods for proof store queries and aggregation

### LINKED ISSUE #176: Start wiring up GRPC implementation
See https://github.com/celestiaorg/celestia-zkevm-hl-testnet/pull/165#pullrequestreview-3180867327

blasrodri:
Add comprehensive gRPC API for querying stored proofs from the ev-prover
service, enabling retrieval of block proofs, membership proofs, and range
proofs from RocksDB storage.

closes #176

### DIFF
--- a/crates/ev-prover/src/config/config.rs
+++ b/crates/ev-prover/src/config/config.rs
@@ -18,7 +18,6 @@ pub struct Config {
     pub evm_rpc: String,
     pub namespace_hex: String,
     pub pub_key: String,
-    pub proof_storage_path: Option<String>,
     #[serde(default = "default_queue_capacity")]
     pub queue_capacity: usize,
     #[serde(default = "default_concurrency")]
@@ -41,7 +40,6 @@ impl Default for Config {
             evm_rpc: "http://127.0.0.1:8545".to_string(),
             namespace_hex: DEFAULT_NAMESPACE.to_string(),
             pub_key: DEFAULT_PUB_KEY_HEX.to_string(),
-            proof_storage_path: None,
             queue_capacity: default_queue_capacity(),
             concurrency: default_concurrency(),
         }
--- a/crates/ev-prover/src/prover/service.rs
+++ b/crates/ev-prover/src/prover/service.rs
@@ -1,10 +1,10 @@
 #![allow(dead_code)]
 use anyhow::Result;
 use std::sync::Arc;
-use storage::proofs::{ProofStorage, RocksDbProofStorage};
+use storage::proofs::ProofStorage;
 use tonic::{Request, Response, Status};
 
-use crate::config::config::{Config, APP_HOME};
+use crate::config::config::Config;
 use crate::proto::celestia::prover::v1::prover_server::Prover;
 use crate::proto::celestia::prover::v1::{
     BlockProof, GetBlockProofRequest, GetBlockProofResponse, GetBlockProofsInRangeRequest,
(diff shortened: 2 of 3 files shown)


crates/ev-prover/src/prover/service.rs
```diff
@@ -1,49 +1,217 @@
...
-    ProveStateTransitionResponse,
+    BlockProof, GetBlockProofRequest, GetBlockProofResponse, GetBlockProofsInRangeRequest,
+    GetBlockProofsInRangeResponse, GetLatestBlockProofRequest, GetLatestBlockProofResponse,
+    GetLatestMembershipProofRequest, GetLatestMembershipProofResponse, GetMembershipProofRequest,
+    GetMembershipProofResponse, GetRangeProofsRequest, GetRangeProofsResponse, MembershipProof, RangeProof,
 };
 use crate::prover::programs::range::BlockRangeExecProver;
 
 pub struct ProverService {
     block_range_prover: BlockRangeExecProver,
```
damiannolan: If you want, you can remove this field now as it won't be needed here 🙏🏻

crates/ev-prover/src/bin/query_proofs.rs
damiannolan: I think you can remove this binary file. I think a better follow up issue after we merge this would be having query CLIs build into the ev-prover binary. 

From example, if I'm running `ev-prover start` in one shell, or even on a remote server. It should be possible for me to open another and run something like this for example.

```
ev-prover query block-proof 100
```

This CLI command would simply spin up a grpc client and make an rpc call either to a remote or locally running node. 

I would suggest we turn this comment into an issue and work on it in a follow up PR, you're more than welcome to pick it up if you would like!

crates/e2e/src/bin/query_stored_proofs.rs
damiannolan: What do you think about just relying on unit tests for this instead. I appreciate the intension to have e2e testing, but it relies on having a node running (with proofs in the store).

I was thinking that because we have a `trait` abstraction for the storage backend, that it should be straight forward to have a `MockProofStorage` implementation, which just stores things in simple map or something - basically an in memory db. 

And that we could leverage that for more unit style testing which would cover this. Would be totally fine with adding the `MockProofStorage` and unit tests in a follow PR to this one if you want to get this merged quicker. Let me know what you think

hyperlane/cmd/hyp/cmd/broadcaster.go
```diff
@@ -21,13 +22,23 @@ import (
...
-	chainID   = "celestia-zkevm-testnet"
 	denom     = "utia"
 	feeAmount = 800
 	gasLimit  = 200000
 )
 
+var (
+	mnemonic = getEnvOrDefault("HYP_MNEMONIC", "sphere exhibit essay fancy okay tuna leaf culture elbow drum trip exchange scorpion excuse parent sun make spot chunk mouse tenant shoe hurt scale")
+	chainID  = getEnvOrDefault("HYP_CHAIN_ID", "celestia-zkevm-testnet")
+)
```
damiannolan: These seem a bit outside the scope of the PR, but I definitely support these changes regardless so whatever 👍🏻

crates/ev-prover/src/prover/programs/block.rs
```diff
@@ -395,6 +395,7 @@ impl BlockExecProver {
                             error!("prove failed: {e:#}");
                         }
                     });
+                    info!("prove generated a proof");
```
damiannolan: want to remove this log before we merge?

crates/ev-prover/src/config/config.rs
```diff
@@ -18,6 +18,7 @@ pub struct Config {
     pub evm_rpc: String,
     pub namespace_hex: String,
     pub pub_key: String,
+    pub proof_storage_path: Option<String>,
```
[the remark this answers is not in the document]
damiannolan: Right, but can't we just construct that proofs db inside `server/mod.rs` and wrap it in an Arc passing to each. 

I done that in the PR I have in progress here -> https://github.com/celestiaorg/celestia-zkevm-hl-testnet/pull/244/files#diff-9eb5ad5d60937e4ae7ff31f668f5ce37cf6d0577c9d92f8ea9b76b98e2081537R37-R43

Because I'm sharing the same proofs db storage object between `BlockExecProver` and `BlockRangeProver`

damiannolan (approved):
Feel free to merge when ready. Idk if we need the storage option or not, either way we can revisit if needs be.

Great work!

blasrodri:
> Feel free to merge when ready. Idk if we need the storage option or not, either way we can revisit if needs be.
> 
> Great work!

I followed your comment with this, and removed that `Option`

PR #251, opened 2025-10-16
Fix: Add Trusted Celestia Root and Height to RangeExecProver Outputs

jonas089:
Please see https://github.com/celestiaorg/celestia-app/pull/6037.

This PR adds the previous celestia root and height to the outputs of the range exec prover, such that they can be verified in the ISM. This is necessary to ensure that no DA blocks were skipped.

### DIFF
--- a/build-local.sh
+++ b/build-local.sh
@@ -1,1 +1,0 @@
-docker build -t ghcr.io/celestiaorg/hyperlane-init:local -f hyperlane/Dockerfile .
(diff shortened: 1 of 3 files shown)


build-local.sh
```diff
@@ -0,0 +1 @@
+docker build -t ghcr.io/celestiaorg/hyperlane-init:local -f hyperlane/Dockerfile .
```
damiannolan: If you want you can just add this to the makefile as:

`make docker-build-hyperlane`

PR #259, opened 2025-10-23
Feat: Combined Circuit for Economical proving in SP1

jonas089:
This PR introduces an alternative mode for `ev-prover` that uses a circuit which batches blocks and therefore minimizes the overhead from generating multiple proofs. 

The main motivation behind this is to save PROVE tokens when using the network, because of the base fee.
In theory the individual block proofs mode (default) allows for faster finality because it has a higher degree of parallelization. 

Parallelization is advantageous when fast finality is required and large blocks occur, but benefits may vary depending on workload and dedicated prover hardware.

For a network with 1 minute finality we can use the new `combined` mode. In order to switch just re-install ev-prover:

```bash
cargo uninstall ev-prover
cargo install --path crates/ev-prover --features combined
```

Then run:

```bash
make start
```
wait for the containers to initialize fully and start the prover service:

```bash
ev-prover init
RUST_LOG="ev_prover=debug" ev-prover start
```

### DIFF
--- a/crates/sp1/ev-combined/README.md
+++ b/crates/sp1/ev-combined/README.md
@@ -1,1 +1,26 @@
 ## Overview
+
+### Program Inputs
+| Name | Type | Description |
+|---|---|---|
+| blocks | Vec<BlockExecInput> | Inputs required to execute a range of blocks |
+
+
+Note that by design proofs are written separately in recursive circuits like this one.
+We use the standard SP1 approach to write compressed proofs generated using `ev-exec`.
+
+### Program Outputs
+| Name | Type | Description |
+|---|---|---|
+| prev_celestia_header_hash | [u8;32] | the trusted Celestia header hash in the ISM |
+| prev_celestia_height | u64 | the trusted Celestia height in the ISM |
+| celestia_header_hash | [u8;32] | the new Celestia header hash after applying the blocks |
+| celestia_height | u64 | the new Celestia height after applying the blocks |
+| trusted_height | u64 | the trusted EV height in the ISM | 
+| trusted_state_root | [u8;32] | the trusted EV state root in the ISM |
+| new_height | u64 | the new EV height after applying the blocks |
+| new_state_root | [u8;32] | the new EV state root after applying the blocks |
+| namespace | [u8;29] | the Celestia namespace that contains the data blobs which themselves contain EV blocks | |
+| public_key | [u8;32] | the sequencer's public key for verification |
+
+These are the same as for `ev-range-exec`, because the circuits fulfill a similar purpose and update the same ISM endpoint. The only difference is that `combined` does not prove every single block recursively, but instead takes a block range as input and generates a single proof using just one prover instance.
--- a/crates/sp1/ev-exec/README.md
+++ b/crates/sp1/ev-exec/README.md
@@ -3,28 +3,37 @@
 An SP1 program that verifies inclusion of EVM reth blocks in the Celestia data availability network 
 and executes their state transition functions.
 
-### Program Inputs
 
-- `CelestiaHeader`: A celestia block header at height H.
-- `DAH`: The associated data availability header at height H.
-- `Namespace`: The namespace containing blob data.
-- `PublicKey`: The public key of the sequencer signing blob data.
-- `Blobs`: All blobs in the namespace for the current block.
-- `NamespaceProofs`: Namespaced Merkle Tree proofs for the complete namespace.
-- `EthClientExecutorInputs`: List of RSP based EVM state transition functions for N blocks included at height H.
-- `TrustedHeight`: A trusted height containing the trusted state root.
-- `TrustedStateRoot`: A trusted state root for the trusted height.
+### Program Inputs
+| Name | Type | Description |
+|---|---|---|
+| header_raw | Vec<u8> | the Celestia block header |
+| dah | DataAvailabilityHeader | the Celestia DA header |
+| blobs_raw | Vec<u8> | The bytes of the data blobs |
+| pub_key | Vec<u8> | The sequencer pubkey for verification |
+| namespace | Namespace | the Celestia namespace that contains the data blobs which themselves contain EV blocks |
+| proofs | Vec<NamespaceProof> | Merkle Proofs for the Namespace data, can be exclusion |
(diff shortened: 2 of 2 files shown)


crates/ev-prover/Cargo.toml
```diff
@@ -13,6 +16,7 @@ anyhow = "1.0"
 async-trait = "0.1.88"
 bincode = { workspace = true }
 bytes = { workspace = true }
+celestia-client = "0.2.0"
```
damiannolan: Interesting... this looks like a new crate? 

> It combines the functionality of [celestia-rpc](https://github.com/eigerco/lumina/blob/HEAD/client/celestia_rpc) and [celestia-grpc](https://github.com/eigerco/lumina/blob/HEAD/client/celestia_grpc) crates.

crates/ev-prover/src/prover/programs/combined.rs
```diff
@@ -0,0 +1,373 @@
...
+    }
+
+    pub fn default_config(prover: &SP1Prover) -> CombinedProverConfig {
+        let (pk, vk) = prover.setup(EV_COMBINED_ELF);
+        CombinedProverConfig::new(pk, vk, SP1ProofMode::Groth16)
+    }
+
+    pub async fn run(self, ism_client: Arc<CelestiaIsmClient>, is_proving_messages: Arc<Mutex<bool>>) -> Result<()> {
+        let client = Client::new(&self.app.celestia_rpc, None).await?;
+        let sequencer_rpc_url = std::env::var("SEQUENCER_RPC_URL").expect("SEQUENCER_RPC_URL must be set");
```
damiannolan: we should probably add this to the config.yaml but also follow up is fine

crates/ev-prover/Cargo.toml
jonas089: Yeah unused, I confused celestia-rpc and celestia-client somehow.

damiannolan: Resolving :)

crates/ev-prover/src/lib.rs
```diff
@@ -20,6 +20,8 @@ use rsp_rpc_db::RpcDb;
 use std::{fs, sync::Arc};
 use tracing::debug;
 
+pub const ISM_ID: &str = "0x726f757465725f69736d000000000000000000000000002a0000000000000001";
```
damiannolan: In the range prover PR the other day I added an ism_id to the client. Which is configured as the "default ism" of the client. This is done via env var since the client is constructed with `from_env`. 

See: https://github.com/celestiaorg/celestia-zkevm-hl-testnet/blob/main/crates/celestia-grpc-client/src/client.rs#L66-L69

And: https://github.com/celestiaorg/celestia-zkevm-hl-testnet/blob/main/.env.example#L16

This should be able to be removed in favour of that API on the client.
Maybe you can address this one before merge.

damiannolan (approved):
Awesome work getting this working in such a short space of time. 
I will create issues for many of the follow up tasks so that we can start getting smaller well scoped PRs in to address some of the comments in here etc..

I left one more which I think we should do before merge! 

Thank you @jonas089 🙏🏻

crates/sp1/ev-combined/program/src/main.rs
```diff
@@ -0,0 +1,26 @@
...
+//!
+//! It performs:
+//! 1. Proof verification for each input
+//! 2. Sequential header verification (i.e., block continuity)
+//! 3. Aggregation of metadata into a `EvmRangeExecOutput`
+//!
+//! It commits:
+//! - The trusted block height and state root
+//! - The new block height and state root
+//! - The latest Celestia header hash from the sequence
```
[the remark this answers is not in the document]
damiannolan:
```suggestion
//! An SP1 program that verifies the execution and inclusion of ev-reth blocks in
//! the Celestia data availability network.
//!
//! This program combines the functionality contained within `ev-exec` and `ev-range-exec`.
//!
//! It accepts N BlockExecInputs which include:
//! - Celestia block header and associated data availability header (DAH).
//! - Namespace
//! - Blobs
//! - Sequencer Public Key
//! - NamespaceProofs
//! - EthClientExecutorInputs (RSP - state transition function)
//! - Trusted Height
//! - Trusted State Root
//!
//! For each Celestia block it performs the following steps:
//! 1. Deserializes the program inputs.
//! 2. Verifies completeness of the namespace using the provided blobs.
//! 3. Executes the EVM blocks via the state transition function.
//! 4. Filters blobs to SignedData and verifies the sequencer signature.
//! 5. Verifies equivalency between the EVM block data and blob data via SignedData.
//!
//! The program then verifies sequentiality of each Celestia block output, ensuring a contiguous series.
//!
//! It commits:
//! - The trusted Celestia block height and header hash.
//! - The new Celestia block height and header hash.
//! - The trusted ev-reth block height and state root
//! - The new ev-reth block height and state root
//! - The Namespace and public key of the sequencer.
```

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.88.0"
+channel = "nightly"
```
damiannolan: Why are we changing this to nightly? 

Shouldn't we stay on 1.88.0 stable

jonas089: yes absolutely, did that by mistake when running udeps.

crates/ev-prover/src/commands/command.rs
```diff
@@ -108,7 +108,7 @@ pub async fn create_zkism() -> Result<()> {
     let pub_key = get_sequencer_pubkey(sequencer_rpc_url).await?;
 
     let ev_hyperlane_vkey = fs::read("testdata/vkeys/ev-hyperlane-vkey-hash")?;
-    let ev_combined_vkey = fs::read("testdata/vkeys/ev-range-exec-vkey-hash")?;
+    let ev_combined_vkey = fs::read("testdata/vkeys/ev-combined-vkey-hash")?;
```
damiannolan: We should just setup a cpu prover in the create cmd and use it to `setup(ELF)` for these. 
Using the local file paths is brittle because it expects the binary to be invoked from a location which has `testdata/vkeys/...` on their machine. 

We should also load the config from yaml in this function rather than using env variables.

I will create an issue and open a PR

PR #264, opened 2025-10-28
Fix: Deploy and Update ZKISM using Rust CLI

jonas089:
There was a tricky issue with the deployment using the Go hyp client, so instead we have to do this in Rust now. 

The celestia height must be chosen such that it contains at least one blob and the trusted state must be initialized to exactly the last blob from the trusted Celestia block. This PR solves this issue and aligns them perfectly.

When running the E2E or prover, we now have to deploy the ISM and Update using the ev-prover CLI. This is not yet automated.

Deployment command:

```
RUST_LOG="ev_prover=info" cargo run -p ev-prover create
```

Update command:
```
RUST_LOG="ev_prover=info" cargo run -p ev-prover update 0x726f757465725f69736d000000000000000000000000002a0000000000000001 0x726f757465725f61707000000000000000000000000000010000000000000000
```

### DIFF
--- a/crates/e2e/src/bin/e2e.rs
+++ b/crates/e2e/src/bin/e2e.rs
@@ -28,8 +28,6 @@ const RETRY_DELAY: u64 = 2;
 #[tokio::main]
 #[allow(clippy::field_reassign_with_default)]
 async fn main() {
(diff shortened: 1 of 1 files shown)


damiannolan (approved):
Approving to unblock you, we can clean things up later 👍🏻

PR #266, opened 2025-10-29
feat: add validator announce, multisig ISM deployment, and local-doma…

blasrodri:
…in parameter

This commit combines two feature additions to the Hyperlane CLI tool:

1. Validator Announce and Multisig ISM Deployment:
   - Added deployment command for MerkleRootMultisigIsm with validator addresses and threshold
   - Added validator announcement command for signature storage location
   - Added helper functions for parsing ISM IDs and validators from events

2. Local Domain Parameter:
   - Added local-domain as a configurable parameter for deployment commands
   - Updated SetupWithIsm to accept dynamic local-domain instead of hardcoded value
   - Applied to deploy-zkism, deploy-noopism, and deploy-multisigism commands

### DIFF
--- a/hyperlane/README.md
+++ b/hyperlane/README.md
@@ -65,8 +65,17 @@ hyperlane warp deploy --config ./configs/warp-config.yaml --registry ./registry 
 ```
 go install ./cmd/hyp
 
-hyp deploy 127.0.0.1:9090
+# Deploy with NoopISM (for testing)
+hyp deploy-noopism 127.0.0.1:9090 69420
+
+# Deploy with ZK Execution ISM
+hyp deploy-zkism 127.0.0.1:9090 [evm-rpc] [ev-node-rpc] 69420
+
+# Deploy with Multisig ISM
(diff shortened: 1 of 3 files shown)


hyperlane/README.md
```diff
@@ -65,9 +65,18 @@ hyperlane warp deploy --config ./configs/warp-config.yaml --registry ./registry
...
 
-hyp deploy 127.0.0.1:9090
+# Deploy with NoopISM (for testing)
+hyp deploy-noopism 127.0.0.1:9090 69420
+
+# Deploy with ZK Execution ISM
+hyp deploy-zkism 127.0.0.1:9090 [evm-rpc] [ev-node-rpc] 69420
+
+# Deploy with Multisig ISM
+hyp deploy-multisigism 127.0.0.1:9090 [validators-comma-separated] [threshold] 69420
```
damiannolan: Maybe it would be good to add here that the validators list should be eth-style 20 byte 0x hex addresses.

blasrodri:
> In rust? You mean Go, right?

No. @jonas089 referred to the integration of some of these commands onto the `ev-prover`

PR #289, opened 2025-11-03
Feat: Prover Server Struct that spawns tasks

jonas089:
This PR adds a prover Server struct that can be used to spawn tasks. The lifecycle management includes joining these tasks and handling them gracefully. Moving forward we will have to make it such that if one tasks fails we re-start the service automatically.

For the `combined` mode I introduced a handler task that will automatically re-start both the combined and message task if either of them fail. That means we are protected from most random errors like request timeouts or malformed responses.

### DIFF
--- a/crates/ev-prover/src/server/mod.rs
+++ b/crates/ev-prover/src/server/mod.rs
@@ -155,17 +155,18 @@ pub async fn start_server(config: Config) -> Result<()> {
     let mut config_clone = config.clone();
     config_clone.pub_key = public_key(sequencer_rpc_url).await?;
     debug!("Successfully got pubkey from evnode: {}", config_clone.pub_key);
-    let client_config = ClientConfig::from_env()?;
-    let client = CelestiaIsmClient::new(client_config).await?;
     // Initialize RocksDB storage in the default data directory
     let storage_path = Config::storage_path().join("proofs.db");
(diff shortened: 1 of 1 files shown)


crates/ev-prover/src/server/mod.rs
```diff
@@ -31,138 +32,258 @@ use crate::prover::service::ProverService;
...
+                        combined_handle.abort();
+                    }
+                }
+                tokio::time::sleep(Duration::from_secs(1)).await;
             }
-        }
-    });
+        });
+
+        wrapper_handle.await?;
```
blasrodri: isn't this an infinite loop? this means nothing below would run, right?

blasrodri: is this the expected behavior?

crates/ev-prover/src/server/mod.rs
```diff
@@ -31,138 +32,258 @@ use crate::prover::service::ProverService;
...
+            r = &mut message_handle => {
+                error!("message prover stopped: {:?}", r);
+                block_handle.abort();
+                block_range_handle.abort();
             }
-        });
+            r = &mut block_range_handle => {
+                error!("block range prover stopped: {:?}", r);
+                block_handle.abort();
+                message_handle.abort();
```
blasrodri: so you mean we abort, and re-initialize? where is the re-initialization happening?

PR #291, opened 2025-11-03
refactor: encapsulate db filenames in storage constructors

### LINKED ISSUE #276: Join db path in db constructor functions
The `Config` object now contains a method `storage_path()` to return the prover service base data directory (e.g. `~/.ev-prover/data`). 

We should provide this base storage path to the db constructors and then join with `proofs.db`, `messages.db`..etc 

> IMO we should pass in the storage path and let the db impl `.join("proofs.db")`. Same for messages db

_Originally posted by @damiannolan in https://github.com/celestiaorg/celestia-zkevm-hl-testnet/pull/262#discussion_r2477466009_

blasrodri:
Pass base storage path to database constructors and let each implementation join with its specific filename (proofs.db, messages.db, snapshots.db) internally.

Closes #276

### DIFF
--- a/crates/ev-prover/tests/integration.rs
+++ b/crates/ev-prover/tests/integration.rs
@@ -30,9 +30,9 @@ async fn test_run_message_prover() {
     let tmp = TempDir::new().expect("cannot create temp directory");
     let storage_path = tmp.path();
 
-    let hyperlane_message_store = Arc::new(HyperlaneMessageStore::new(&storage_path).unwrap());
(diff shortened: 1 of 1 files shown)


crates/ev-prover/src/server/mod.rs
damiannolan: There's a function called `prepare_message_prover` in this file which is appending the `snapshots.db` and `messages.db` suffixes still.

damiannolan (changes requested):

PR #294, opened 2025-11-07
Deps: Update EV-Reth and and EV-Node to the latest release tags as of Nov 7 2025

jonas089:
This PR updates our docker environment to use the latest release tags for ev-reth and ev-node.

Moving forward we should include the tag version as a comment alongside the sha tag from the docker build, because it is practically impossible to find the corresponding rev or release tag given just the sha hash of the image.


jonas089:
> Thanks for handling this. This addresses the long awaited #76.
> 
> I assume you have tested this successfully?

@damiannolan I would like to test it one more time before converting it from draft to ready for review. Today I'm working on the steps required to make our internal testnet public.

PR #302, opened 2025-11-19
Fix: Limit Indexer Range to 100k

jonas089:
@damiannolan As we discussed, this PR fixes the max range exceeded error in the indexer, by batching max. 100k blocks at a time.

I made the max range configurable the same way that we currently configure the batch sizes.

### DIFF
--- a/crates/ev-prover/src/prover/programs/batch.rs
+++ b/crates/ev-prover/src/prover/programs/batch.rs
@@ -3,7 +3,7 @@ use std::time::{Duration, Instant};
 
 use crate::prover::abi::MailboxContract;
 use crate::prover::chain::ChainContext;
-use crate::prover::config::MAX_INDEXING_RANGE;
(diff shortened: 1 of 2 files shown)


crates/ev-prover/src/prover/programs/batch.rs
```diff
@@ -205,19 +206,21 @@ impl BatchExecProver {
             let (proof, output) = self.prove(input).await?;
             info!("Proof generation time: {}", start_time.elapsed().as_millis());
 
-            // index if new ev blocks were included
-            if status.trusted_height < output.new_height {
+            // index if new ev blocks were included, the maximum range that reth supports by default is 100000 blocks
+            let mut start = status.trusted_height + 1;
+            while start <= output.new_height {
```
damiannolan: So if its an empty block proof the loop will not run 👍🏻 Logic checks out!

PR #310, opened 2025-11-25
feat: Generic ISM for any State Transition

jonas089:
This PR goes hand-in-hand with https://github.com/celestiaorg/celestia-app/tree/feature/generic-ism, where Ethereum consensus proofs (SP1 Helios) were added to the ZKISM. All changes are related to the proto types.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,6 @@
 services:
   celestia-init:
-    image: ghcr.io/celestiaorg/celestia-app-standalone:feature-zk-execution-ism
+    image: ghcr.io/celestiaorg/celestia-app-standalone@sha256:ea0e11656c638bfdd62bf2541a3cf580304925c3e00917a337033ff2062da4cb
     container_name: celestia-init
     entrypoint: /scripts/init.sh
     volumes:
@@ -9,7 +9,7 @@ services:
     restart: "no"
 
   celestia-validator:
-    image: ghcr.io/celestiaorg/celestia-app-standalone:feature-zk-execution-ism
+    image: ghcr.io/celestiaorg/celestia-app-standalone@sha256:ea0e11656c638bfdd62bf2541a3cf580304925c3e00917a337033ff2062da4cb
     container_name: celestia-validator
     ports:
       - "26656:26656"


crates/ev-zkevm-types/src/programs/block.rs
```diff
@@ -95,47 +99,131 @@ pub struct BatchExecInput {
...
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Debug)]
 pub struct BlockRangeExecOutput {
-    // prev_celestia_header_hash is the merkle hash of the previous Celestia block header.
-    pub prev_celestia_header_hash: [u8; 32],
-    // prev_celestia_height is the height of the previous Celestia block.
-    pub prev_celestia_height: u64,
-    // celestia_header_hash is the hash of the celestia header at which new_height is available.
+    // the length prefix of the state
+    pub state_len_le_u64_bytes: [u8; 8],
```
damiannolan: I don't know if we need to mention little endian u64 bytes in the field names..
I would be fine with that being in the comments above the fields. 

Something like this might look a bit cleaner to work with, bit of a nit, but ¯\_(ツ)_/¯
```rust
// state length prefix (little-endian u64)
pub state_len: [u8; 8],
// trusted previous state
pub state: State,
// new state length prefix (little-endian u64)
pub new_state_len: [u8; 8],
// the result of the state transition
pub new_state: State,
```

crates/ev-zkevm-types/src/programs/block.rs
```diff
@@ -95,47 +99,131 @@ pub struct BatchExecInput {
...
+    // the starting point of the state transition
+    pub state: State,
+    // the length prefix of the new state
+    pub new_state_len_le_u64_bytes: [u8; 8],
+    // the result of the state transition
+    pub new_state: State,
+}
+
+impl Serialize for BlockRangeExecOutput {
```
damiannolan: Are these implementations being used? It looks like below on line 537 we manually encode rather than construct an output and call serialize on it

crates/sp1/ev-range-exec/program/src/main.rs
```diff
@@ -108,5 +117,14 @@ pub fn main() {
...
+    let state_serialized = bincode::serialize(&state).expect("failed to serialize state");
+    let new_state_serialized = bincode::serialize(&new_state).expect("failed to serialize new state");
+    let state_len_le_64_bytes = (state_serialized.len() as u64).to_le_bytes().to_vec();
+
+    let mut output: Vec<u8> = Vec::new();
+    output.extend_from_slice(&state_len_le_64_bytes);
+    output.extend_from_slice(&state_serialized);
+    output.extend_from_slice(&new_state_serialized);
+
+    sp1_zkvm::io::commit_slice(&output);
```
damiannolan: Couldn't we construct the `BlockRangeExecOutput` and call `serialize()` on it?

crates/ev-zkevm-types/src/programs/block.rs
jonas089: Agreed,  a comment is good enough.

crates/sp1/ev-range-exec/program/src/main.rs
damiannolan: ah wait, sorry that doesn't add up since the output should be the `BlockRangeExecOutput` struct itself, but under the hood sp1 libs will call serialize on it which would call the serializer trait impl

jonas089: BlockRangeExecOutput is fine, we can use serialize but like I mentioned above we might have to keep my custom serializer impl to ensure it's always treated as fixed-size.

crates/ev-zkevm-types/src/programs/block.rs
jonas089: Okay it's not necessary with the current types, so I removed the custom serializer.

crates/ev-zkevm-types/src/programs/block.rs
```diff
@@ -458,3 +483,89 @@ impl BlockVerifier {
...
+        assert_eq!(output.new_state_len_bytes, deserialized.new_state_len_bytes);
+        assert_eq!(output.new_state.state_root, deserialized.new_state.state_root);
+        assert_eq!(
+            output.new_state.celestia_header_hash,
+            deserialized.new_state.celestia_header_hash
+        );
+        assert_eq!(output.new_state.celestia_height, deserialized.new_state.celestia_height);
+        assert_eq!(output.new_state.height, deserialized.new_state.height);
+        assert_eq!(output.new_state.namespace, deserialized.new_state.namespace);
+        assert_eq!(output.new_state.public_key, deserialized.new_state.public_key);
```
damiannolan: I'm pretty sure that we can replace this with a single `assert_eq!()` if we derive `PartialEq, Eq` on both `State` and `BlockRangeExecOutput`

crates/sp1/ev-batch-exec/README.md
```diff
@@ -9,18 +9,24 @@
...
 | public_key | [u8;32] | the sequencer's public key for verification |
 
-These are the same as for `ev-range-exec`, because the circuits fulfill a similar purpose and update the same ISM endpoint. The only difference is that `batch-exec` does not prove every single block recursively, but instead takes a block range as input and generates a single proof using just one prover instance.
\ No newline at end of file
+### BlockRangeExecOutput: Circuit outputs used for on-chain verification
+| Name | Type | Description |
+|---|---|---|
+| state_len_bytes | [u8;8] | Little endian encoded bytes of the length of the State used by this program |
+| state | State | The starting point of the state transition
+| new_state | State | The result of the state transition
```
damiannolan: I think we need to add the other length prefix here

damiannolan (approved):
Awesome job! 🙏🏻 
Super nice to get this all done before the end of the week.

PR #322, opened 2025-12-02
feat: Manual Retry for Hyperlane Script

jonas089:
This PR adds manual snapshot and message submission retry to our Hyperlane circuit CLI (in script).

Planned follow ups:
- http server that exposes snapshot finality for monitoring (work in progress)
- automated snapshot submission (not started)
- pruning of finalized snapshots (not started)


jonas089:
We should merge https://github.com/celestiaorg/celestia-zkevm/pull/326 and create a tag before we merge this PR.

crates/sp1/ev-hyperlane/script/Cargo.toml
```diff
@@ -12,17 +12,22 @@ path = "src/bin/main.rs"
 name = "vkey-ev-hyperlane"
 path = "src/bin/vkey.rs"
 
+[features]
+retry = []
```
damiannolan: Idk if we need the feature flag for this on this binary, do we? It seems like it just adds a bit more operational overhead for having to compile things correctly etc. 

I think we could just include this without the feature flag. What do you think?

damiannolan (approved):
Happy to go ahead with the separate binary and consider additions to ev-prover later.

I left a comment suggesting that we remove the feature flags tho. Let me know what you think. Otherwise approved 🙏🏻

PR #333, opened 2025-12-19
fix: Celestia Light Client for Header Root Verification with Generic Verifier

jonas089:
We removed the header root verification from the verifier in order to make it more generic.

Therefore we now have to ensure the header correctness against the active validator set (chain of trust) inside our circuit.

This PR adds tendermint consensus verification to the batched circuit.

Planned release for this change: 0.2.1 (non-breaking with respect to the ISM).

### DIFF
--- a/crates/ev-zkevm-types/Cargo.toml
+++ b/crates/ev-zkevm-types/Cargo.toml
@@ -25,5 +25,4 @@ serde = { workspace = true }
 serde_cbor = { workspace = true }
 sha3 = { workspace = true }
 tendermint = { workspace = true }
-
 tendermint-light-client-verifier = { workspace = true }


crates/ev-prover/src/config/mod.rs
```diff
@@ -164,6 +164,9 @@ pub struct RpcConfig {
     /// RPC endpoint for the Celestia light node.
     pub celestia_rpc: String,
 
+    /// RPC endpoint for the Tendermint light node.
```
damiannolan:
```suggestion
    /// RPC endpoint for the consensus node.
```

damiannolan (commented):
Overall looks great. Thanks @jonas089 

I know its still in draft but I dropped some comments since I was taking a look.

crates/ev-zkevm-types/Cargo.toml
```diff
@@ -25,3 +25,5 @@ serde = { workspace = true }
 serde_cbor = { workspace = true }
 sha3 = { workspace = true }
 tendermint = { workspace = true }
+
```
damiannolan:
```suggestion
```

damiannolan (approved):
LGTM

PR #345, opened 2026-01-09
Feat: TEE prover service implementation

### LINKED ISSUE #340: Feat: TEE prover service
Add an implementation of the prover service that leverages the TEE attestations from https://github.com/celestiaorg/evolve-tee.

jonas089:
This PR adds a prover service implementation for the TEE service. 
A new feature flag `tee_mode` was introduced alongside `batch_mode` that will start the TEE prover service + message prover service, similar to how we did it for the batch prover.

Detailed documentation can be found in `TEE.md`.


crates/e2e/src/config.rs
damiannolan: fwiw I think we should delete the e2e directory in this repo now? I think its kind of legacy code right? and is slightly unmaintained. I'm not sure it actually works rn, and I think we've both been manually testing with the prover service.

Perhaps we could add a tastora test since we now have hyperlane deployment support there. Let's discuss it outside of this PR

crates/ev-prover/src/prover/programs/tee.rs
```diff
@@ -0,0 +1,346 @@
...
+            proof.public_values.as_slice(),
+        )?)
+    }
+
+    fn prover(&self) -> Arc<SP1Prover> {
+        Arc::clone(&self.prover)
+    }
+}
+
+impl TeeExecProver {
```
damiannolan: Am I correct in thinking that much of this impl is the same as the batched prover?

hyperlane/relayer-config.json
```diff
@@ -33,7 +33,7 @@
             "interchainSecurityModule": "0xa05915fD6E32A1AA7E67d800164CaCB12487142d",
             "mailbox": "0xb1c938F5BA4B3593377F399e12175e8db0C787Ff",
             "merkleTreeHook": "0x1D957dA7A6988f5a9d2D2454637B4B7fea0Aeea5",
-            "protocolFee": "0xFCb1d485ef46344029D9E8A7925925e146B3430E",
+            "protocolFee": "0x1D957dA7A6988f5a9d2D2454637B4B7fea0Aeea5",
```
damiannolan: is this change correct? I know it shouldn't technically affect the relayer but both addresses for merkle tree hook and protocol fee hook are the same here. 

Relayer doesn't care about protocl fee hook (afaik)

hyperlane/configs/core-config.yaml
```diff
@@ -1,5 +1,5 @@
 defaultHook:
-  address: "0xFCb1d485ef46344029D9E8A7925925e146B3430E"
```
damiannolan: Same here, I'm not sure that this should change ?

damiannolan (commented):
Nice work, in terms of the prover service integration looks clean enough to me, but I need to get a better understanding of the flow overall.

crates/ev-prover/src/config/mod.rs
```diff
@@ -70,6 +70,20 @@ impl Config {
     /// The groth16 verifier key.
     const GROTH16_VK: &[u8] = include_bytes!("../../resources/groth16_vk.bin");
 
+    /// Creates a Config instance from environment variables.
+    pub fn from_env() -> Result<Self> {
+        Ok(Self {
+            grpc_address: "127.0.0.1:50051".into(),
+            rpc: RpcConfig::from_env()?,
+            hyperlane: HyperlaneConfig::default(),
```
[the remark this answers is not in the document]
damiannolan: Same comment here, should this be using `from_env()`?

There was a `from_env` impl added for `CelestiaHyperlaneConfig` but I don't see where its being used.

crates/ev-prover/src/prover/programs/batch.rs
```diff
@@ -147,7 +144,8 @@ impl BatchExecProver {
 
     /// Returns the prover config.
     pub fn default_config(prover: &SP1Prover) -> BatchProverConfig {
-        let (pk, vk) = prover.setup(BATCH_ELF);
+        let batch_elf = std::fs::read("elfs/ev-batch-elf").expect("Failed to read ev-batch-elf");
```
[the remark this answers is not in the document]
damiannolan: Same comment here is still applicable I think. If I run the binary `ev-prover` from anywhere on my machine, then I would need to ensure I have `elfs/` directory in the correct location, correct?

crates/ev-prover/src/prover/programs/message.rs
```diff
@@ -134,7 +131,8 @@ impl HyperlaneMessageProver {
 
     /// Returns the default prover configuration for the block execution program.
     pub fn default_config(prover: &SP1Prover) -> MessageProverConfig {
-        let (pk, vk) = prover.setup(EV_HYPERLANE_ELF);
+        let ev_hyperlane_elf = std::fs::read("elfs/ev-hyperlane-elf").expect("Failed to read ev-hyperlane-elf");
```
damiannolan: Same applies here

crates/ev-prover/src/prover/programs/tee.rs
```diff
@@ -0,0 +1,512 @@
...
+use serde::{Deserialize, Serialize};
+use sp1_sdk::{SP1ProofMode, SP1ProofWithPublicValues, SP1ProvingKey, SP1Stdin, SP1VerifyingKey};
+use storage::hyperlane::message::HyperlaneMessageStore;
+use tee_attestation_types::{AttestationResponse, Inputs as TeeAttestationInput};
+
+use tokio::{sync::mpsc, time::interval};
+use tracing::{debug, error, info, warn};
+
+#[derive(Deserialize, Serialize)]
+struct AttestationRequest {
```
damiannolan, line 36: nit: can we move this struct after the `use crate` definition below

damiannolan: How come we are using hex strings over the wire? Is that the best approach?

crates/ev-prover/TEE.md
```diff
@@ -0,0 +1,157 @@
...
+
+> **⚠️ WARNING: Without these additional constraints the TEE circuit is NOT SAFE TO USE IN PRODUCTION!**
+
+## Overview
+
+The TEE integration provides hardware-based security guarantees for the zkEVM prover. The architecture is designed for minimal dependencies:
+
+- **Prover**: Fetches block inputs and light blocks from Celestia/EVM nodes
+- **TEE App**: Receives serialized data via POST, performs native verification, generates attestation
+- **No middleware required**: The prover communicates directly with the TEE app
```
damiannolan: We can remove this line

hyperlane/configs/core-config.yaml
damiannolan: This comment still is applicable. This address is being changed for some reason to the same address of the `requiredHook` below, which is a merkle tree hok.

hyperlane/relayer-config.json
damiannolan: Same here, this should not change

.env.example
```diff
@@ -32,7 +32,18 @@ CELESTIA_NAMESPACE="a8045f161bf468bf4d44"
...
-RETH_RPC_URL="http://localhost:8545"
+RETH_RPC_URL="http://127.0.0.1:8545"
 RETH_WS_URL="ws://127.0.0.1:8546"
-CELESTIA_RPC_URL="http://localhost:26658"
-SEQUENCER_RPC_URL="http://localhost:7331"
+SEQUENCER_RPC_URL="http://127.0.0.1:7331"
+CELESTIA_RPC_URL="http://127.0.0.1:26658"
+TENDERMINT_RPC_URL="http://127.0.0.1:26657"
+CELESTIA_GRPC_ENDPOINT="http://127.0.0.1:9090"
+MIDDLEWARE_ENDPOINT="http://127.0.0.1:9091"
```
damiannolan: We can remove this

damiannolan (commented):
Prover impl looks good! Nice work. 

There's still some comments that need to be addressed before we merge tho.
- Configuration (see comments on `from_env`)
- Reading ELFs from expected relative directory
- Hyperlane contract address changes
- Remove any dangling references to "middleware"

hyperlane/configs/core-config.yaml
damiannolan: Can we put this back before merge

hyperlane/relayer-config.json
damiannolan: And here too!

damiannolan (approved):
Approving! Feel free to merge after the contract addresses are put back for protocol fee hook!

jonas089:
> 0xFCb1d485ef46344029D9E8A7925925e146B3430E

These are the same as the MerkleTreeHook one though, which has changed. I think they're correct?

PR #354, opened 2026-03-17
chore: Reduce duplication and remove deadcode

jonas089:
- Keep only the core prover service logic and make batched the default mode
- Reduce duplication
- Remove parallel mode


.claude/settings.json
damiannolan, line 1: I think we can remove this file, or add to gitignore

crates/ev-prover/src/prover/programs/batch.rs
```diff
@@ -143,119 +71,17 @@ impl BatchExecProver {
     }
 
     /// Returns the prover config.
-    pub fn default_config(prover: &SP1Prover) -> BatchProverConfig {
+    pub fn default_config(prover: &SP1Prover) -> StandardProverConfig {
         let batch_elf = include_bytes!("../../../../../elfs/ev-batch-elf");
```
damiannolan: this could potentially be made as a static const because its included in the binary twice - see in `tests/storage.rs`

damiannolan (commented):
looks pretty good

crates/ev-prover/src/prover/programs/common.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    Ok(BATCH_SIZE)
+}
+
+/// Queries and stores Hyperlane mailbox events from the provided block range (inclusive),
+/// chunking requests to respect `MAX_INDEXING_RANGE`.
+/// The `MAX_INDEXING_RANGE` const is set to align with the default value of 100,000 blocks.
+/// This setting can be configured via the EVM execution client using `max_blocks_per_filter: u64` and `max_logs_per_response: usize`.
+pub async fn index_messages(
```
damiannolan, line 111: maybe its boomer of me to expect these functions to be associated with some type but fine to keep them as standalone for now 👍🏻

damiannolan (approved):
LGTM

PR #356, opened 2026-03-20
refactor: use hyperlane cli only for docker testnet deployments

damiannolan:
## Overview

- Simplify hyperlane deployment
- Move hyperlane registry into `testnet` dir
- Remove cross archx builds to speed up docker workflow builds in CI


damiannolan:
The CodeQL ci is failing because there is no longer any Golang sources in the repository. I have updated github configuration but its pending, so this is just a false positive

Makefile
```diff
@@ -93,13 +93,9 @@ spamoor:
...
 .PHONY: spamoor
 
-e2e:
-	cargo run --bin e2e -p e2e --release
-.PHONY: e2e
-
 docker-build-hyperlane:
 	@echo "--> Building hyperlane-init image"
-	@docker build -t ghcr.io/celestiaorg/hyperlane-init:local -f hyperlane/Dockerfile .
+	@docker build --platform linux/amd64 -t ghcr.io/celestiaorg/hyperlane-init:local -f testnet/hyperlane/Dockerfile .
```
jonas089, line 98: Is there a reason why this is platform specific now?

testnet/hyperlane/scripts/docker-entrypoint.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# HYP_KEY is the priv key of the EVM account used for Hyperlane contract deployment
+export HYP_KEY=0x82bfcfadbf1712f6550d8d2c00a39f05b33ec78939d0167be2a737d691f33a6a
+export HYP_KEY_COSMOSNATIVE=0x6e30efb1d3ebd30d1ba08c8d5fc9b190e08394009dc1dd787a69e60c33288a8c
+
+echo "Using Hyperlane registry:"
+hyperlane registry list --registry ./registry
+
+echo "Deploying Hyperlane core on Evolve..."
+hyperlane core deploy --chain evolve --config ./configs/evolve-core.yaml --registry ./registry --yes
```
jonas089, line 24: Much cleaner. Nice work :)

jonas089 (approved):

Makefile
damiannolan: yes, because I removed the cross architecture builds in the docker workflow, in order to run this on mac arm chips we need to pass this flag

damiannolan: docker builds will be much quicker in CI

PR #357, opened 2026-06-08
feat: Local Proving + Version Bump to V6

jonas089:
This PR introduces local `cuda` proving support.

Additionally it updates the SP1 prover version to 6.2.3 and depends on https://github.com/celestiaorg/celestia-app/pull/7374.

For now the celestia-app image is set to `local`. It should be updated once 7374 is merged.


crates/ev-prover/src/prover/programs/batch.rs
```diff
@@ -1,3 +1,4 @@
+use std::fs::write;
```
damiannolan: unused import here

crates/ev-prover/src/prover/programs/message.rs
```diff
@@ -17,7 +17,8 @@ use ev_zkevm_types::hyperlane::{
     io::{HyperlaneMessageInputs, HyperlaneMessageOutputs},
     proof::{HyperlaneBranchProof, HyperlaneBranchProofInputs, HYPERLANE_MERKLE_TREE_KEYS},
 };
-use sp1_sdk::{SP1ProofMode, SP1ProofWithPublicValues, SP1Stdin};
+use sp1_sdk::{Elf, Prover, ProvingKey, SP1ProofMode, SP1ProofWithPublicValues, SP1Stdin};
+use std::fs::write;
```
damiannolan: ditto

testnet/celestia-app/Dockerfile
damiannolan, line 1: Does the PR on celestia-app build a docker image tagged like `pr-xyz` with PR number?

is this just a custom dockerfile to use your branch?

damiannolan (approved):
Just left a question about docker image used, and some unused imports. 
Would be preferable to use an image tag built from celestia-app but fine if we want to temporarily use a custom one

crates/ev-prover/src/prover/programs/batch.rs
jonas089: Ah right that was from when I generated the test fixture.

testnet/celestia-app/Dockerfile
jonas089: For now it's just a local build of https://github.com/celestiaorg/celestia-app/pull/7374. Let me check if I can find the docker build in CI and link it.

jonas089: It's been a while since we last did that and forgot that the CI builds it so should be able to add it hopefully.

